PR #8, opened 2020-08-27
feat(client): introduce Socky 🧦 - the nifty internal socket state manager

enisdenjo:
Socky aims to introduce "concurrency" safety. Allows for sinks to subscribe/dispose in parallel while maintaining the WebSocket connection without races.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -251,9 +251,15 @@ export function createClient(options: ClientOptions): Client {
       }
       subscribedSinks[uuid] = sink;
 
-      let messageListener: Disposable | undefined;
+      let messageListener: Disposable | undefined,
+        completed = false;
       prepare()
         .then(() => {
+          // the sink might have completed before the socket became ready
+          if (completed) {
+            return;
+          }
+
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -33,6 +33,110 @@ export interface Client extends Disposable {
...
+      if (state.connecting || state.disconnecting) {
+        return await this.shouldConnect();
+      }
+
+      return !state.connected;
+    },
+    connecting() {
+      state = { connecting: true, connected: false, disconnecting: false };
+    },
+    connected(connectedSocket: Socket) {
```
enisdenjo, line 84: When connected is called, you can only pass an instantiated WebSocket.

src/client.ts
```diff
@@ -33,6 +33,110 @@ export interface Client extends Disposable {
   subscribe<T = unknown>(payload: SubscribePayload, sink: Sink<T>): () => void;
 }
 
+/** The nifty internal socket state manager: Socky 🧦. */
+function createSocky() {
+  type Socket = WebSocket | null;
```
enisdenjo, line 38: Superfluous type.

PR #19, opened 2020-09-29
GraphQL Over WebSocket New PR GIF

amshalem:


enisdenjo (approved):
Absolutely brilliant! Thank you very much @amshalem.

I have just one more request - could you please be kind and send me the logo in a vector format? Would be great if you can include the bridge in the background so that I update the socials of the repo and elsewhere. Thanks in advance!

PR #28, opened 2020-10-17
feat(server): Support returning multiple results from `execute`

n1ru4l:
This PR adds supporting the changes of https://github.com/graphql/graphql-js/pull/2319 ([but also for allowing to use graphql-transport-ws with my live query implementation](https://github.com/n1ru4l/graphql-live-queries/blob/a4ce5ac727a7c44fd205b824d678db7717abd220/packages/in-memory-live-query-store/src/InMemoryLiveQueryStore.ts#L96-L103))

@enisdenjo Please let me know your opinion on adding this 😄

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -75,7 +75,7 @@ Direction: **Client -> Server**
 
 Requests an operation specified in the message `payload`. This message provides a unique ID field to connect future server messages to the operation started by this message.
 
-If there is already an active subscriber for a `subscription` operation matching the provided ID, the server will close the socket immediately with the event `4409: Subscriber for <unique-operation-id> already exists`. Since `query` and `mutation` resolve to a single emitted value, their subscription does not require reservations for additional future events. Having this in mind, the server may not assert this rule for these operations.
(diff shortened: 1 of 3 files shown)


enisdenjo:
Thank you for this PR! Before I dig in reviewing the code, I'd just like to touch on one point - shouldn't live queries be requested through the `subscription` operation?

For example, if you want to make this live:
```gql
{
  hello
}
```
you'd do:
```gql
subscription {
  hello
}
```

I am not so familiar with live queries, as I never used them 😅, but, this is [how PostGraphile](https://www.graphile.org/postgraphile/live-queries/) does it. Even in the [Hasura live queries architecture document](https://github.com/hasura/graphql-engine/blob/master/architecture/live-queries.md), they send live query events through the subscriptions operation.

n1ru4l:
I recommend reading https://medium.com/open-graphql/graphql-subscriptions-vs-live-queries-e38302c7ab8e

There is no official spec for live queries. The drawback of using subscriptions is that you can only query for one root field and need to also add all your query root fields to the subscription root type. Aside from that both live queries and subscriptions can be used to do the same things.

enisdenjo:
Thats fair! Thanks for sharing your insights and the great article. I learned something new. :D

enisdenjo (commented):
This looks great! Thank you for contributing with this PR.

We probably have to change a few details the protocol document as well. So I'll glance on that and merge altogether then.

enisdenjo (approved):

Issue #29, opened 2020-10-17
tests are failing on windows and LTS node v12.19.0


n1ru4l:
I recently got a windows machine and the tests are failing when executed.

```
FAIL  src/tests/server.ts
  ● Connect › should not close the socket after the `connectionInitWaitTimeout` has passed but the callback is still resolving

    expect.assertions(2)

    Expected two assertions to be called but received one assertion call.

      268 | 
      269 |   it('should not close the socket after the `connectionInitWaitTimeout` has passed but the callback is still resolving', async () => {
    > 270 |     expect.assertions(2);
          |            ^
      271 | 
      272 |     await makeServer({
      273 |       connectionInitWaitTimeout: 10,

      at Object.<anonymous> (src/tests/server.ts:270:12)

  ● Connect › should close the socket if an additional `ConnectionInit` message is received while one is pending

    expect.assertions(3)

    Expected three assertions to be called but received four assertion calls.

      298 | 
      299 |   it('should close the socket if an additional `ConnectionInit` message is received while one is pending', async () => {
    > 300 |     expect.assertions(3);
          |            ^
      301 | 
      302 |     await makeServer({
      303 |       connectionInitWaitTimeout: 10,

      at Object.<anonymous> (src/tests/server.ts:300:12)

 FAIL  src/tests/client.ts
  ● "concurrency" › should dispatch and receive messages even if one subscriber disposes while another one subscribes

    expect(jest.fn()).toBeCalled()

    Expected number of calls: >= 1
    Received number of calls:    0

      399 |     expect(nextFnForHappy).not.toBeCalled();
      400 |     expect(completeFnForHappy).toBeCalled();
    > 401 |     expect(nextFnForBananas).toBeCalled();
          |                              ^
      402 |   });
      403 | });
      404 | 

      at Object.<anonymous> (src/tests/client.ts:401:30)

  ● lazy › should connect immediately when mode is disabled

    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 2

      411 |     await wait(10);
      412 | 
    > 413 |     expect(server.webSocketServer.clients.size).toBe(1);
          |                                                 ^
      414 |     server.webSocketServer.clients.forEach((client) => {
      415 |       expect(client.readyState).toBe(WebSocket.OPEN);
      416 |     });

      at Object.<anonymous> (src/tests/client.ts:413:49)

Test Suites: 2 failed, 2 total
[...]

PR #32, opened 2020-10-20
fix(tests): Avoid relying on setTimeout where possible.

n1ru4l:
This PR attempts to solve https://github.com/enisdenjo/graphql-transport-ws/issues/29 for the server test suite.

### DIFF
--- a/src/tests/server.ts
+++ b/src/tests/server.ts
@@ -859,7 +859,7 @@ describe('Subscribe', () => {
   it('should execute the subscription and "next" the published payload', async () => {
     expect.assertions(1);
 
-    await makeServer({
+    const [server] = await makeServer({
       schema,
     });
 
@@ -878,6 +878,14 @@ describe('Subscribe', () => {
       const message = parseMessage(data);
       switch (message.type) {
         case MessageType.ConnectionAck: {
+          onReceiveNextMessage(server, () => {
+            pubsub.publish('becameHappy', {
+              becameHappy: {
+                name: 'john',
+              },
+            });
+          });
+
           client.send(
             stringifyMessage<MessageType.Subscribe>({
               id: '1',
@@ -891,16 +899,6 @@ describe('Subscribe', () => {
                 }`,
               },
             }),
-            () =>
-              setTimeout(
-                () =>
-                  pubsub.publish('becameHappy', {
-                    becameHappy: {
-                      name: 'john',
-                    },
-                  }),
-                0,
-              ),
           );
           break;
         }
@@ -1073,9 +1071,7 @@ describe('Subscribe', () => {
 
 describe('Keep-Alive', () => {
   it('should dispatch pings after the timeout has passed', async () => {
-    expect.assertions(1);
-    jest.useFakeTimers();
-
+    expect.assertions(0);
     const [server] = await makeServer({
       keepAlive: 50,
     });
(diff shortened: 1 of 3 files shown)


enisdenjo:
🚀  Thanks for taking a look into this! I will review it as soon as I can.

In the meantime, please add the `Closes #29` in the description of the PR to close the referenced issue automatically when the merge lands.

src/tests/server.ts
```diff
@@ -854,51 +956,60 @@ describe('Subscribe', () => {
...
     // confirm complete
-    expect(onMessageFn).toHaveLastReturnedWith({
+    expect(value).toEqual({
       id: '1',
       type: MessageType.Complete,
     });
 
-    pubsub.publish('boughtBananas', new Error('Something weird happened!'));
-    await wait(10);
+    // TODO: if we use something like a PushPull publish we can better confirm nobody is subscribing anymore
```
n1ru4l, line 1000: See https://github.com/n1ru4l/graphql-live-queries/blob/a4ce5ac727a7c44fd205b824d678db7717abd220/packages/in-memory-live-query-store/src/PushPullAsyncIterableIterator.ts#L1

We could simply check `pullQueue.length` 🤔

But since the idea of the test is that it would show that no more messages should be dispatched it is also hard to test this without waiting 😔

src/tests/server.ts
```diff
@@ -962,83 +1073,103 @@ describe('Subscribe', () => {
 
 describe('Keep-Alive', () => {
   it('should dispatch pings after the timeout has passed', async () => {
-    await makeServer({
+    expect.assertions(1);
+    jest.useFakeTimers();
```
n1ru4l, line 1077: This is also hard to test I tried `jest.advanceTimersByTime(49)` in order to prove that the ping is not sent earlier. However then I had to add multiple `await new Promise(setImmediate);` afterwards instead of a single one for the `callback` to be invoked.

The event loop timings with fake timers behave incomprehensible to me...

n1ru4l: Maybe it is easier to just skip the fake timers and just prove that the ping is invoked.

sibelius:
everything is running better now

I've just got this warning

```jsx
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.
```

n1ru4l:
I also still need to check the client test suite.

n1ru4l:
> everything is running better now
> 
> I've just got this warning
> 
> ```js-jsx
> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.
> ```

This was caused by the fake timers. I removed them and increased the timeouts for the test cases where we need to use setTimeout. https://github.com/enisdenjo/graphql-transport-ws/pull/32/commits/38c5e8eb0a2b842432d2040663760db740bc483c

enisdenjo (commented):
This looks awesome and I am very grateful for your help! But, in my honest opinion, solutions to the problems this attempts to solve can be done in a much simpler and more efficient fashion. All while increasing tests clarity and extensibility.

Taking inspiration from your PR, I have made an attempt on tackling this on my own in #33. Its still a draft, but cleaning up the leftovers shouldn't take long. Would be awesome if you can take a look at it, I'd like to hear your opinion!

I hope you don't mind me taking over after you did this work already 😬.

n1ru4l:
No problem!

PR #33, opened 2020-10-20
refactor(server): Stabilize and simplify tests

### LINKED ISSUE #29: tests are failing on windows and LTS node v12.19.0

enisdenjo:
Closes #29, closes #32

A few takeaways:
- Guaranteed shutdown of the server between tests. I've been naively simply using the `server.close()` function which actually [stops the server from accepting new connections and keeps existing connections](https://nodejs.org/api/net.html#net_server_close_callback)
- Super simple testing client

### DIFF
--- a/src/tests/fixtures/simple.ts
+++ b/src/tests/fixtures/simple.ts
@@ -6,6 +6,7 @@ import {
   subscribe,
   GraphQLNonNull,
 } from 'graphql';
+import net from 'net';
 import http from 'http';
 import { PubSub } from 'graphql-subscriptions';
 import { createServer, ServerOptions, Server } from '../../server';
@@ -80,10 +81,17 @@ export const port = 8273,
 
 export async function startServer(
   options: Partial<ServerOptions> = {},
-): Promise<[Server, () => Promise<void>]> {
+): Promise<[Server, (beNice?: boolean) => Promise<void>]> {
   const httpServer = http.createServer((_req, res) => {
     res.writeHead(404);
     res.end();
+  });
+
+  // sockets to kick off on teardown
+  const sockets = new Set<net.Socket>();
+  httpServer.on('connection', (socket) => {
+    sockets.add(socket);
+    httpServer.once('close', () => sockets.delete(socket));
   });
 
   const server = await createServer(
@@ -103,14 +111,19 @@ export async function startServer(
 
   return [
     server,
-    () =>
+    (beNice) =>
       new Promise((resolve, reject) => {
-        const disposing = server.dispose();
-        if (disposing instanceof Promise) {
-          disposing.catch(reject).then(() => {
-            httpServer.close((err) => (err ? reject(err) : resolve()));
-          });
+        if (!beNice) {
+          for (const socket of sockets) {
+            socket.destroy();
+            sockets.delete(socket);
+          }
         }
+
+        const disposing = server.dispose() as Promise<void>;
+        disposing.catch(reject).then(() => {
+          httpServer.close(() => resolve());
+        });
       }),
   ];
 }
--- a/src/tests/fixtures/simple.ts
+++ b/src/tests/fixtures/simple.ts
@@ -13,6 +13,17 @@ import { createServer, ServerOptions, Server } from '../../server';
 
 export const pubsub = new PubSub();
 
+// use for dispatching a `pong` to the `ping` subscription
+let pendingPongs = 0;
+let nextPong: ((done: boolean) => void) | undefined;
+export function pong(): void {
+  if (nextPong) {
+    nextPong(false);
+  } else {
+    pendingPongs++;
+  }
+}
+
 const personType = new GraphQLObjectType({
   name: 'Person',
   fields: {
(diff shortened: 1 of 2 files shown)


n1ru4l:
I like this approach!

I stress tested it by hitting up watch mode and triggering rebuilds manually with Enter and I still hit some timeouts e.g.

```bash
 FAIL  src/tests/server.ts (5.619 s)
  ● Subscribe › should execute the subscription and "next" the published payload

    : Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error:

      848 |   });
      849 | 
    > 850 |   it('should execute the subscription and "next" the published payload', async () => {
          |   ^
      851 |     await makeServer({
      852 |       schema,
      853 |     });

      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (src/tests/server.ts:850:3)
```

enisdenjo:
I am glad you like it! 😁

Hmm, funny, as I've been stress testing it the same way on 2 different mac machines - passing every single time.

Taking another glance at the test in focus - it might be possible that the `SubscribeMessage` hadnt settled before dispatching the first event through `pubsub.publish`. As there is no "subscription confirmed" message from the server, I'll dig through the alternatives.

Thanks for the heads up!

enisdenjo:
After short investigation, I've noticed weird behaviour in `graphql-subscriptions` PubSub system.

Even after adding my attempt on [waiting for subscription settlement](https://github.com/enisdenjo/graphql-transport-ws/pull/33/commits/e9b310bf212ea5a1fadc377c088996230dccadca), the test will fail if you call `pubsub.publish` serially after the subscription went through (uncomment the `setTimeout` to see it in action).

n1ru4l:
Isn't this a timing issue? `client.waitForMessage` must be called before emitting the values.

enisdenjo:
It is not a timing issue, the `waitForMessage` uses a dispatch queue:
https://github.com/enisdenjo/graphql-transport-ws/blob/e9b310bf212ea5a1fadc377c088996230dccadca/src/tests/server.ts#L51-L70

enisdenjo:
Rebased.

enisdenjo:
Rebased.

sibelius:
one thing that we could improve is to have 1 server per `it` so we could run tests in parallel, and also make sure that 1 test does not influence the other test

n1ru4l:
> one thing that we could improve is to have 1 server per it so we could run tests in parallel, and also make sure that 1 test does not influence the other test

Arent tests inside a single failed always run in sequence?

enisdenjo:
All unit tests are run serially within a suite. To prevent mixing in the client and the server, I also run separate suites serially thanks to [jest's -i/--runInBand argument](https://github.com/enisdenjo/graphql-transport-ws/blob/fcaaa351f1c3e7d32e2356fbdb5554717216548d/package.json#L52)

After each test, the server is shut down:
https://github.com/enisdenjo/graphql-transport-ws/blob/13459f2d7b460667d616e190820e52c0f66ad908/src/tests/server.ts#L21-L27

Before each test, the server is spun up again (notice the `await makeServer();` on the beginning of every test).

Running tests in parallel is not straight-forward because every server will have its own address, I think running tests serially is OK for now (They still finish within 10 seconds in the CI 😉).

sibelius:
tests are running well and faster for me

sibelius:
supertest enable running tests in parallel

it has a helper function like this:

```jsx
export const serverAddress = (server: Server) => {
  const addr = server.address();

  if (!addr) {
    server.listen(0);
  }

  const port = server.address().port;
  const path = '';

  return `localhost:${port}${path}`;
};
```

so for each test it will be a different address and port

and then you use like this

```jsx
const server = http.createServer(app.callback());

const address = serverAddress(server);

const ws = new WebSocket(`ws://${address}/ws`);

```

enisdenjo:
Nice!

I am still wary about adding this complication simply because premature optimization is the root of all evil. But, I'll keep this in mind for when the tests scale and parallelism becomes a necessity.

n1ru4l:
I opened https://github.com/enisdenjo/graphql-transport-ws/pull/35 for the approach that skips pubsub and simply uses a AsyncIterableIterator that we can publish the values with. Let me know what you think.

enisdenjo:
Rebased.

PR #35, opened 2020-10-21
fix: Use custom AsyncIterableIterator for the publish tests

n1ru4l:
Without this change `this` inside the return method of the iterator will be undefined. It also solves the ugly pubsub hacks by using a AsyncIterableIterator that is returned from subscribe. We can then simply publish values with `.push()` and also check whether `return` was called on the iterator.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -324,7 +324,7 @@ export function createServer(
 
       Object.entries(ctxRef.current.subscriptions).forEach(
         ([, subscription]) => {
-          (subscription.return || noop)();
+          subscription.return?.();
         },
       );
     }
@@ -544,7 +544,7 @@ export function createServer(
           }
           case MessageType.Complete: {
             if (ctx.subscriptions[message.id]) {
-              await (ctx.subscriptions[message.id].return ?? noop)();
+              await ctx.subscriptions[message.id].return?.();
             }
             break;
           }
--- a/src/tests/utilities/push-pull-async-iterable-iterator.ts
+++ b/src/tests/utilities/push-pull-async-iterable-iterator.ts
@@ -1,0 +1,57 @@
+export class PushPullAsyncIterableIterator<T>
+  implements AsyncIterableIterator<T> {
+  public pushQueue: T[] = [];
+  private pullQueue: ((value: IteratorResult<T>) => void)[] = [];
+  private isRunning = true;
+
+  public async next(): Promise<IteratorResult<T>> {
+    return new Promise((resolve) => {
+      if (this.pushQueue.length !== 0) {
+        resolve(
+          this.isRunning
+            ? // We check for length
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              { value: this.pushQueue.shift()!, done: false }
+            : { value: undefined, done: true },
+        );
+      } else {
+        this.pullQueue.push(resolve);
+      }
+    });
+  }
+
+  public async return(): Promise<IteratorResult<T>> {
+    if (this.isRunning) {
+      this.isRunning = false;
+      for (const resolve of this.pullQueue) {
+        resolve({ value: undefined, done: true });
+      }
+      this.pullQueue.length = 0;
+      this.pushQueue.length = 0;
+    }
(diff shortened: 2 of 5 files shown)


src/server.ts
```diff
@@ -324,7 +324,7 @@ export function createServer(
 
       Object.entries(ctxRef.current.subscriptions).forEach(
         ([, subscription]) => {
-          (subscription.return || noop)();
+          subscription.return?.();
```
n1ru4l, line 327: The former implementation resolved in `this` being undefined inside a class method.

src/server.ts
```diff
@@ -544,7 +544,7 @@ export function createServer(
           }
           case MessageType.Complete: {
             if (ctx.subscriptions[message.id]) {
-              await (ctx.subscriptions[message.id].return ?? noop)();
+              await ctx.subscriptions[message.id].return?.();
```
n1ru4l, line 547: Same as above

src/tests/utilities/push-pull-async-iterable-iterator.ts
```diff
@@ -0,0 +1,57 @@
+export class PushPullAsyncIterableIterator<T>
+  implements AsyncIterableIterator<T> {
+  public pushQueue: T[] = [];
+  private pullQueue: ((value: IteratorResult<T>) => void)[] = [];
+  private isRunning = true;
```
n1ru4l, line 5: This AsyncIterable will add pending items to the push queue and pending next call promises to the pull queue. New values can be added by calling `push`. This allows us more control over what is actually happening where PubSub is a black box we have no control over. It prevents us from using hacks like `setTimeout` in order to fix the timings.

enisdenjo (commented):
While I appreciate your work and help a lot, this is just over-engineering it. There are no queues, message pressure, network latency, etc. in tests - they should be stupid simple. Sorry to say, but IMHO, `PushPullAsyncIterableIterator` is just an overkill and an extra thing that has to be maintained.

Even `graphql-subscriptions` is an overkill... I realised that I ended up testing its functionality over focusing on our own. I settled with [an ultra simple ping/pong dispatcher](https://github.com/enisdenjo/graphql-transport-ws/pull/33/commits/7e3150d42efd0b3774e9ea8a59a592b4283505e0). When I come by refactoring the client tests, I will use it there too and eventually completely drop the `graphql-subscriptions` development dependency.

Anything not related to the WebSockets is just superfluous. And, as you may notice - I don't like code, I like _no_ code. "Everything should be made as simple as possible, but no simpler."

src/server.ts
enisdenjo: I don't really understand what you mean by this. Can you elaborate a bit more please?

enisdenjo: Nevertheless, I like the looks of this much better! Since this PR is closed as part of #33, could you please make a separate PR with these changes?

n1ru4l:
It's fine with me, I see your current solution also works, this was just the first thing I had in mind given I had worked on it before 😄 

I actually disagree with latency. Since a real websocket connection is established there is always latency. However, this was nothing the code ever intended to solve 😅

I would still appreciate if you could apply the changes for the return call so that library consumers are not forced to explicitly bind the this scope to the return function/method. Also it looks cleaner in the code than the noop call.

enisdenjo:
You're right! I spoke too quickly about latency while trying to push a point. 😄

enisdenjo:
And to be honest, I didnt know that you can do nullish coalescing on function calls! I tried with `some.func?()` but TS kept complaining, you showed me the way with the dot (`some.func?.()`) - thanks!

n1ru4l:
Yeah, the syntax is a bit weird 😄  

No problem I am just glad there is a library for subscriptions where someone is responding to issues 🙏

enisdenjo:
> I would still appreciate if you could apply the changes for the return call so that library consumers are not forced to explicitly bind the this scope to the return function/method. Also it looks cleaner in the code than the noop call.

This just landed in [v1.8.2](https://github.com/enisdenjo/graphql-transport-ws/releases/tag/v1.8.2).

Issue #36, opened 2020-10-22
documentation of how to handle Authorization


sibelius:
### Story

I'm want to be able to send user authorization token that can modify the GraphQL context

---
When using `subscriptions-transport-ws` I could generate the GraphQL Context using the `onConnect` helper

https://github.com/sibelius/relay-workshop/blob/master/packages/server/src/index.ts#L30

```jsx
import { SubscriptionServer } from 'subscriptions-transport-ws';

SubscriptionServer.create(
    {
      onConnect: async (connectionParams: ConnectionParams) => {
        const { user } = await getUser(connectionParams?.authorization);

        return getContext({ user });
      },
      // eslint-disable-next-line
      onDisconnect: () => console.log('Client subscription disconnected!'),
      execute,
      subscribe,
      schema,
    },
    {
      server,
      path: '/subscriptions',
    },
  );
```

When using this package, we need to pass `context` on server creating that we won't be able to modify context based on use authorization token

enisdenjo:
You can simply use the [`onSubscribe`](https://github.com/enisdenjo/graphql-transport-ws/blob/master/docs/interfaces/_server_.serveroptions.md#onsubscribe) and inject your context before GraphQL execution. Here's how:
```ts
import { createServer } from 'graphql-transport-ws';

createServer(
  {
    schema,
    execute,
    subscribe,
    onSubscribe: async (ctx, _msg, args) => {
      const { user } = await getUser(ctx.connectionParams?.authorization);
      return [{ ...args, contextValue: await getContext({ user }) }];
    },
  },
  {
    server,
    path: '/graphql',
  },
);
```

or if you want to get the context once [`onConnect`](https://github.com/enisdenjo/graphql-transport-ws/blob/master/docs/interfaces/_server_.serveroptions.md#onconnect) and just reuse it, here is how you'd go by:

```ts
import { createServer } from 'graphql-transport-ws';

const authorizedContexts: Record<string, unknown> = {};

createServer(
  {
    schema,
    execute,
    subscribe,
    onConnect: async (ctx) => {
      if (typeof ctx.connectionParams?.authorization !== 'string'){
        return false;
      }

      const { user } = await getUser(ctx.connectionParams?.authorization);
      authorizedContexts[ctx.connectionParams.authorization] = await getContext({ user });
      return true;
    },
    onSubscribe: (ctx, _msg, args) => {
      return [
        {
          ...args,
[...]

PR #46, opened 2020-11-03
feat(server): For dynamic usage, `context` option can be a function too

enisdenjo:
Inspired by https://github.com/enisdenjo/graphql-ws/issues/36#issuecomment-721095047

Helpful for when you want a dynamic context, but you want to avoid using the `onSubscribe` because you have to set the complete execution arguments up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,16 +490,18 @@ server.listen(443);
 </details>
 
 <details>
-<summary>Server usage with custom static GraphQL arguments</summary>
+<summary>Server usage with custom GraphQL context value</summary>
 
 ```typescript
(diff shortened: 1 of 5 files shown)


tautvilas:
I think this would be a great addition. Custom context injection is a very common thing: I want to inject such things as auth, feature toggles or maybe some data from the request (for example IP address of customer). And in all these cases I don't want to be fiddling with GraphQL internals.

enisdenjo:
Exactly, you're arguments make a lot of sense! Am merging this as soon as I write up some tests.

PR #49, opened 2020-11-04
fix(server): Consistently set `rootValue` and `contextValue`, if not overridden

benjie:
Previously we had:

- rootValue is only overwritten if onSubscribe didn't return anything
- context is overwritten if onSubscribe didn't return anything OR if onSubscribe's return value doesn't state context.

This PR makes this behaviour consistent by replacing the `rootValue` condition with:

- rootValue is overwritten if onSubscribe didn't return anything OR if onSubscribe's return value doesn't state rootValue

It also adjusts the check such that a user returning `contextValue: null` or `rootValue: null` (or anything else falsy) from `onSubscribe` will not have those values overwritten.

### DIFF
--- a/docs/interfaces/_server_.serveroptions.md
+++ b/docs/interfaces/_server_.serveroptions.md
@@ -204,10 +204,11 @@ If you return `ExecutionArgs` from the callback,
 it will be used instead of trying to build one
 internally. In this case, you are responsible
 for providing a ready set of arguments which will
-be directly plugged in the operation execution. Beware,
-the `context` server option is an exception. Only if you
-dont provide a context alongside the returned value
-here, the `context` server option will be used instead.
+be directly plugged in the operation execution.
+
+Omitting the fields `contextValue` or `rootValue`
+from the returned value will have the provided server
+options fill in the gaps.
 
 To report GraphQL errors simply return an array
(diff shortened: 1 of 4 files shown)


n1ru4l:
hey there, I am watching this repository and having the feeling that the number of the configuration option combinations is getting quite complex.

I recently implemented my own transport layer over Socket.io and decided to only allow providing rootValue/contextValue/schema and all other relevant parameters via one factory function that is invoked for each incoming operation execution request. ([You can find it here](https://github.com/n1ru4l/graphql-live-queries/blob/db24ff8d254544ff6be3cfd4985acc50970f1f2b/packages/socket-io-graphql-server/src/registerSocketIOGraphQLServer.ts#L28-L72))

Such a method would allow implementing all those combinations this library addresses in user-and thus reducing complexity.

enisdenjo:
You can still provide all necessary GraphQL arguments through a single source - [the `onSubscribe` callback](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/_server_.serveroptions.md#onsubscribe). This PR helps in cases where the user decides to mix the various inputs in.

enisdenjo (approved):

PR #51, opened 2020-11-04
refactor: tweak types and checks for timeouts/intervals

benjie:
When reviewing this, you might want to turn on the ignore whitespace differences setting.

- Handles the situation where timeouts are negative (which should disable them)
- Uses `isFinite` over `!== Infinity` (also handles NaN, etc)
- If the checks fail, uses `null` as the fallback, rather than `false`.


enisdenjo (approved):
Brilliant!

PR #53, opened 2020-11-04
fix(server): Close socket if `onSubscribe` returns invalid array

benjie:
At the moment an empty array, or an array of things that don't match the error requirements, will be interpreted as `execArgs` which'll then cause all property references to be null and to have confusing errors thrown. This change handles the case where an array is returned but it doesn't match the requirements for an error array.

I'm not sure if emitting the error like this is the best route?


enisdenjo (approved):
Thanks for the catch! Misconfiguring your server is bad business, so I prefer killing the socket in favour of adjusting your implementation to comply with the requirements.

PR #54, opened 2020-11-04
fix: handle error from operationResult stream

benjie:
In general `operationResult` should only stream successes or the end of the stream; however I imagine that there could be a circumstance in which it'd reject, throwing an error from the `for await`; this handles that case. I'm not sure if this is the correct direction; it just occurred to me that there wasn't any local error handling here and maybe there should be?

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -640,11 +640,16 @@ export function createServer(
               }
               ctx.subscriptions[message.id] = operationResult;
 
-              for await (const result of operationResult) {
-                await emit.next(result, execArgs);
+              try {
+                for await (const result of operationResult) {
+                  await emit.next(result, execArgs);
+                }
+                await emit.complete();
+              } catch (err) {
+                await emit.error(err);
+              } finally {
+                delete ctx.subscriptions[message.id];
               }
-              await emit.complete();
-              delete ctx.subscriptions[message.id];
             } else {
               /** single emitted result */
 


src/server.ts
```diff
@@ -640,11 +640,16 @@ export function createServer(
...
 
-              for await (const result of operationResult) {
-                await emit.next(result, execArgs);
+              try {
+                for await (const result of operationResult) {
+                  await emit.next(result, execArgs);
+                }
+                await emit.complete();
+              } catch (err) {
+                await emit.error(err);
```
enisdenjo, line 649: We cannot guarantee here that this error will be an array of `GraphQLError`s. Which the error message requires.
https://github.com/enisdenjo/graphql-ws/blob/1bb9218ad3ee9442442122c1d10910d51951b763/src/message.ts#L54-L58

enisdenjo (commented):
> In general `operationResult` should only stream successes or the end of the stream;

Exactly because of this statement, I left the loop bare. If something throws from here, the higher try/catch will close the socket and clean up subscriptions.

Might this be too harsh, kicking the client off if the execution loop throws?

How often does this happen and what could be the root causes? If misconfiguration is the usual case, then the socket _should_ close, you should tidy up your setup for optimal usage and security.

benjie:
It's not uncommon to have 20+ subscriptions running on a webpage; closing all of those because one misbehaves does feel a little harsh. I guess we should come up with a situation in which the `for await` would throw, and if we can and it's not too far-fetched then merging something like this would make sense. For now I can't think of a way since this is the result of a graphql execute or subscribe and those catch their own errors and wrap them in payloads, so I suspect it's fine. (Just thinking about things like if live queries work in a different way, for example.)

enisdenjo:
Together with the stated arguments and for keeping parity with the existing implementation where throwing an error from _any_ callback closes the socket:
https://github.com/enisdenjo/graphql-ws/blob/b96dbb98bb6f71956321ce1202af0af50df4e40e/src/tests/server.ts#L752-L866
I think its safe to close this PR for now. If this behaviour becomes against expectations, we will readdress this issue.

PR #56, opened 2020-11-04
fix(server): Distribute server error to all clients even if one error listener throws

benjie:
At the moment if the `'error'` handler for one of the clients were to throw then the `'error'` event won't be passed on to any further clients. This is probably not desired? I've wrapped it with a `try/catch` to allow the additional clients to be warned. However, I'm not happy with discarding the error here - we should raise it somehow. We could do something like `let firstError = null`; store the error to firstError in the catch, and then throw if firstError exists... Or we could log... Or... What do you want to do here?

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -445,7 +445,11 @@ export function createServer(
   webSocketServer.on('error', (err) => {
     for (const client of webSocketServer.clients) {
       // report server errors by erroring out all clients with the same error
-      client.emit('error', err);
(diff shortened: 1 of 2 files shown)


enisdenjo (approved):
Ended up doing https://github.com/enisdenjo/graphql-ws/pull/56/commits/a9f0196757e8556cc256b7926126bab610c00c0a 🤷‍♂️.

PR #63, opened 2020-11-11
docs(recipe): add example for apollo-server-express

she11sh0cked:
This adds an example for using apollo-server-express with graphql-ws as a subscription layer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,7 +546,7 @@ const app = express();
 // create apollo server
 const apolloServer = new ApolloServer({
     schema,
-    subscriptions: false // this is important
+    subscriptions: false, // disable subscriptions-transport-ws
 });
 
 // apply middleware
--- a/README.md
+++ b/README.md
@@ -530,6 +530,49 @@ server.listen(443, () => {
 </details>
 
 <details>
+<summary>Server usage with <a href="https://github.com/apollographql/apollo-server/tree/main/packages/apollo-server-express">Apollo Server Express</a></summary>
+
+```typescript
+import https from 'https';
+import express from 'express';
+import { ApolloServer } from 'apollo-server-express';
+import { createServer } from 'graphql-ws';
+import { execute, subscribe } from 'graphql';
+import { schema } from 'my-graphql-schema';
+
+// create express
+const app = express();
+
+// create apollo server
+const apolloServer = new ApolloServer({
+    schema,
+    subscriptions: false // this is important
+});
+
+// apply middleware
+apolloServer.applyMiddleware({ app })
+
+// create a http server using express
+const server = https.createServer(app);
+
+server.listen(443, () => {
+  createServer(
+    {
+      schema,
+      execute,
+      subscribe,
+    },
+    {
+      server,
+      path: '/graphql', // you can use the same path too, just use the `ws` schema
+    },
+  );
+});
+```
+
+</details>
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -529,6 +529,49 @@ server.listen(443, () => {
...
+import { execute, subscribe } from 'graphql';
+import { schema } from 'my-graphql-schema';
+
+// create express
+const app = express();
+
+// create apollo server
+const apolloServer = new ApolloServer({
+    schema,
+    subscriptions: false // this is important
```
enisdenjo, line 549: I guess because it disables `subscriptions-transport-ws`? If so, would be nice to state that.Just to reduce confusion.

```suggestion
    subscriptions: false, // disable subscriptions-transport-ws
```

enisdenjo (changes requested):
Other than the small remark, this looks awesome! Thanks for sharing!

README.md
```diff
@@ -529,6 +529,49 @@ server.listen(443, () => {
...
+import { execute, subscribe } from 'graphql';
+import { schema } from 'my-graphql-schema';
+
+// create express
+const app = express();
+
+// create apollo server
+const apolloServer = new ApolloServer({
+    schema,
+    subscriptions: false, // disable subscriptions-transport-ws
```
kevinwei00, line 549: This is not needed because `(apollo-server-core) installSubscriptionHandlers` is only invoked if you're using `ApolloServer.listen`, Apollo's provided convenience if you don't create your own Express object.

enisdenjo: TBH, am not an Apollo expert... So I cannot confirm this, can you @she11sh0cked? [Seems like `ApolloServer` does something if the `subscriptions` is not `false`.](https://github.com/apollographql/apollo-server/blob/dfd87a0d836f96a5ad8e1538615056799d664f59/packages/apollo-server-core/src/ApolloServer.ts#L289-L323)

@kevinwei00 could you maybe point me to the location in code where this is checked?

kevinwei00: [In this function](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-core/src/ApolloServer.ts#L599), the subscriptionServer is implemented by `subscriptions-transport-ws`. And [here](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server/src/index.ts#L119) is where it gets used (only in `ApolloServer.listen`).

enisdenjo: Gotcha! So only if you're calling the `listen` method on the apollo server the default `subscriptions-transport-ws` will kick in:
```
const apolloServer = new ApolloServer({ ... });
apolloServer.listen(...);
```

enisdenjo: Recipe updated in https://github.com/enisdenjo/graphql-ws/commit/b46a33c4c5ac060e192520399cd2c6afd30a927a.

kevinwei00: That's my understanding, yes.

enisdenjo: Thanks @kevinwei00, I love simplifications! 😄

she11sh0cked: Should we note that in the recipe? Could lead to confusion if someone uses `apolloServer.listen(...);`, right?

enisdenjo: I wouldn't, the recipe explicitly uses the http server. They should be as simple as possible, the MVP for usage.

PR #65, opened 2020-11-12
fix: Stop sending messages after receiving complete

rpastro:
@enisdenjo  I noticed that the client logic was sending a Complete event back to server immediately after the Server had sent a Complete event for that same operation, so hopefully this is a valid change.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -497,7 +497,7 @@ export function createClient(options: ClientOptions): Client {
             // either the canceller will be called and the promise resolved
             // or the socket closed and the promise rejected
             await throwOnCloseOrWaitForCancel(() => {
-              if (!completed && socket.readyState === WebSocketImpl.OPEN) {
+              if (!completed) {
                 // send complete message to server on cancel
                 socket.send(
                   stringifyMessage<MessageType.Complete>({
--- a/src/client.ts
+++ b/src/client.ts
@@ -50,7 +50,7 @@ export type EventListener<E extends Event> = E extends EventConnecting
   ? EventClosedListener
   : never;
 
-type CancellerRef = { current: (() => void) | null };
+type CancellerRef = { current: ((completed: boolean) => void) | null };
 
 /** Configuration used for the `create` client function. */
 export interface ClientOptions {
@@ -210,7 +210,9 @@ export function createClient(options: ClientOptions): Client {
   ): Promise<
     [
       socket: WebSocket,
-      throwOnCloseOrWaitForCancel: (cleanup?: () => void) => Promise<void>,
+      throwOnCloseOrWaitForCancel: (
+        cleanup?: (completed: boolean) => void,
+      ) => Promise<void>,
     ]
   > {
     // prevents too many recursive calls when reavaluating/re-connecting
@@ -248,8 +250,8 @@ export function createClient(options: ClientOptions): Client {
                   return reject(event);
                 }
 
-                cancellerRef.current = () => {
-                  cleanup?.();
+                cancellerRef.current = (completed: boolean) => {
+                  cleanup?.(completed);
                   state.locks--;
                   if (!state.locks) {
                     state.socket?.close(1000, 'Normal Closure');
@@ -374,8 +376,8 @@ export function createClient(options: ClientOptions): Client {
             return reject(event);
           }
 
-          cancellerRef.current = () => {
-            cleanup?.();
+          cancellerRef.current = (completed: boolean) => {
+            cleanup?.(completed);
             state.locks--;
             if (!state.locks) {
               socket.close(1000, 'Normal Closure');
@@ -456,7 +458,7 @@ export function createClient(options: ClientOptions): Client {
               // because you cannot receive a message
               // if there is no existing connection
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              cancellerRef.current!();
+              cancellerRef.current!(false);
               // TODO-db-201025 calling canceller will complete the sink, meaning that both the `error` and `complete` will be
               // called. neither promises or observables care; once they settle, additional calls to the resolvers will be ignored
             }
@@ -468,7 +470,7 @@ export function createClient(options: ClientOptions): Client {
               // because you cannot receive a message
               // if there is no existing connection
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              cancellerRef.current!();
+              cancellerRef.current!(true);
               // calling canceller will complete the sink
             }
             return;
@@ -494,14 +496,16 @@ export function createClient(options: ClientOptions): Client {
 
             // either the canceller will be called and the promise resolved
             // or the socket closed and the promise rejected
-            await throwOnCloseOrWaitForCancel(() => {
-              // send complete message to server on cancel
-              socket.send(
-                stringifyMessage<MessageType.Complete>({
-                  id: id,
-                  type: MessageType.Complete,
-                }),
-              );
+            await throwOnCloseOrWaitForCancel((completed: boolean) => {
+              if (!completed) {
+                // send complete message to server on cancel
+                socket.send(
+                  stringifyMessage<MessageType.Complete>({
+                    id: id,
+                    type: MessageType.Complete,
+                  }),
+                );
+              }
             });
 
             socket.removeEventListener('message', messageListener);
@@ -539,7 +543,7 @@ export function createClient(options: ClientOptions): Client {
         .finally(() => (cancellerRef.current = null)); // when this promise settles there is nothing to cancel
 
       return () => {
-        cancellerRef.current?.();
+        cancellerRef.current?.(false);
       };
     },
     dispose() {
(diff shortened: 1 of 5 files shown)


enisdenjo (changes requested):
Brilliant, thanks for doing this!

However, the same can be achieved with much much less. I would recommend that you refactor so that the `completed` flag is inside the returned `subscribe` method (the subscribe method is the only one that cares...).

Something like this:

```ts
...
subscribe(payload, sink) {
  const id = generateID();
  let completed = false; // 👈 add me
  const cancellerRef: CancellerRef = { current: null };
...
```
flag as completed when receiving message from server
```ts
...
case MessageType.Complete: {
  if (message.id === id) {
    completed = true; // 👈 add me
    // the canceller must be set at this point
    // because you cannot receive a message
    // if there is no existing connection
    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    cancellerRef.current!();
    // calling canceller will complete the sink
  }
  return;
}
...
```
and then, as you did, simply dont send if completed:
```ts
await throwOnCloseOrWaitForCancel(() => {
  if (!completed) { // 👈 check me
    // send complete message to server on cancel
    socket.send(
      stringifyMessage<MessageType.Complete>({
        id: id,
        type: MessageType.Complete,
      }),
    );
  }
});
```

P.S. Would be awesome if you could add a test case for this, if you're willing of course! If not, I'll add one to your PR and then we'll merge - no pressure. 👍

rpastro:
Thanks for the comments. I'll update the PR.

enisdenjo:
BTW, this behaviour was kinda intentional, I've seen it like a "completion" acknowledgement and keeps the code ultra simple. Both the **server** and the **client** do the same, they both will emit a complete message when completed.

If we decide against this on the client side, we must apply necessary changes to the server too!

rpastro:
For our project we are using a Java backend and we have implemented the WS protocol according to the specification, which is nicely written BTW.   Since the spec doesn't describe this behavior (i.e., acknowledging the Complete with another Complete) our server doesn't do that.

I thought about removing the Complete from the client to avoid the extra messages, specially if the client is sending lots of queries and mutations. 

Let me know whether or not you would like to keep this.  If not, I can update the Server side as well. If you would like to keep it we could update the spec. We just would have to decide whether this is optional, recommended or mandatory.

enisdenjo:
Wow, awesome! Lovely to see the Protocol being implemented in other languages too. 😄 

I just remembered that the server does that too, so I left a comment to not forget.

Your PR makes total sense and, to be honest, the extra completed message from either side is simply superfluous. Having this in mind, we should drop it from the library and keep in-sync with the Protocol.

**TL;DR:** Complete completes the stream - no additional messages should be sent through afterwards by either side.

enisdenjo:
Oh, and, the client won't misbehave if you don't send an additional complete message.

rpastro:
@enisdenjo   I pushed a new patch with your suggestions.  I'll take a look at the server side and the unit tests.

src/client.ts
```diff
@@ -495,13 +497,15 @@ export function createClient(options: ClientOptions): Client {
...
             // or the socket closed and the promise rejected
             await throwOnCloseOrWaitForCancel(() => {
-              // send complete message to server on cancel
-              socket.send(
-                stringifyMessage<MessageType.Complete>({
-                  id: id,
-                  type: MessageType.Complete,
-                }),
-              );
+              if (!completed && socket.readyState === WebSocketImpl.OPEN) {
```
rpastro, line 500: I added this check for the socket state because I noticed some errors in my console where socket state was already CLOSING or CLOSED at this point.

enisdenjo, line 500: Why check for the open state?

The cleanup callback is called exclusively if the user cancels the subscription. You can cancel only an active connection. If the socket closes at any point in time, the `throwOnCloseOrWaitForCancel` will throw and the cleanup will never be called.

```suggestion
              if (!completed) {
```

enisdenjo (changes requested):

src/client.ts
enisdenjo: Crazy, I just left a comment about that here: https://github.com/enisdenjo/graphql-ws/pull/65#discussion_r522142673 😅 . This has to be handled then, the socket _should_ be still open for the cleanup.

rpastro: I added this check for the socket state because I noticed some errors in my console where socket state was already CLOSING or CLOSED at this point.

enisdenjo: Resolved in favour of: https://github.com/enisdenjo/graphql-ws/pull/65#discussion_r522140655.

enisdenjo: Okay then, this has be handled! The socket _should_ still be open here. Or maybe my reasoning is wrong. I'll come back to this later...

rpastro: Do you want me to remove the check?

rpastro:
@enisdenjo   I don't see a Complete message from server after client sends a Complete (see attached image).  I also couldn't find anywhere in the code where server would do that.

Screen Shot 2020-11-12 at 9 35 01 AM

enisdenjo:
Interesting. It should be sent... [There is even a test which checks that the server calls the onComplete callback when the client completes](https://github.com/enisdenjo/graphql-ws/blob/ff54d5d84048df6665cf38495bb07fbf36d7a235/src/tests/server.ts#L1564-L1606).

What happens is that, on complete, [the server "returns" the subscription async iterator](https://github.com/enisdenjo/graphql-ws/blob/ff54d5d84048df6665cf38495bb07fbf36d7a235/src/server.ts#L693-L696). This breaks [the streaming loop](https://github.com/enisdenjo/graphql-ws/blob/ff54d5d84048df6665cf38495bb07fbf36d7a235/src/server.ts#L680-L682) and [calls the next line](https://github.com/enisdenjo/graphql-ws/blob/ff54d5d84048df6665cf38495bb07fbf36d7a235/src/server.ts#L683), which, in turn, [emits the complete message](https://github.com/enisdenjo/graphql-ws/blob/ff54d5d84048df6665cf38495bb07fbf36d7a235/src/server.ts#L574-L580).

Its possible that the server is [not sending the message because the socket is not open](https://github.com/enisdenjo/graphql-ws/blob/ff54d5d84048df6665cf38495bb07fbf36d7a235/src/server.ts#L484) (like you said, closing or closed)?

src/client.ts
enisdenjo: I am currently not fully concentrated on this TBH, am swamped by my day job.

But, I will come back to this later investing a proper thought. I hope you understand.

rpastro:
I tried again with multiple subscriptions in parallel to ensure the issue is not related to the socket being closed and I see the same result - i.e., no Complete from server.

Screen Shot 2020-11-12 at 9 48 41 AM

rpastro:
@enisdenjo   I'm so sorry.....  😐    I forgot I'm using our own server implementation.

src/client.ts
rpastro: I fully understand and I apologize for flooding you with message. I'm focused on this exactly because it is overlapping with my day job ATM.  

I'll remove the check as part of this PR  and I'll try debugging it later to see what is happening.

enisdenjo: No worries! I respond when I can. I have unresolved this discussion, because I'd like to get to the bottom of this first.

As mentioned in the https://github.com/enisdenjo/graphql-ws/pull/65#issuecomment-726103670, the client should not complain about the server not sending the complete message. So, you can leave your server implementation be and the client will still work.

**However,** if this is not the case for you - I will be happy to merge just the client fix to elevate your block! In this case, please put the readyState check back. Also, if we merge, I would kindly ask you to open an issue for the server fix too (you can leave the implementation to me).

rpastro: The client doesn't have any problems with the server not sending the Complete message.

I'm making the enhancements to the server side and unit tests. I'll push an updated patch soon.

rpastro: The issue with socket state seems to be unrelated to these changes, so I am planning on investigating that separately.

rpastro:
@enisdenjo  I pushed a new patch for the server updates.  

I also looked at the unit tests for the client, but I couldn't find a way to verify that the client is not sending the Complete message back to the server.  As far as I can tell, the only way to test this would be to expose what is happening at the socket level either at the client or the server, but I didn't want to touch that.

src/client.ts
enisdenjo: Just a short question: you were using the server from this lib (not your own) when the state issues were occurring, right?

enisdenjo (commented):
Hey, thanks for your support! There are elements that I would like to alter, however - I'd be faster doing them myself. So, I'll push a few commits to this PR and wrap it up. 😄

src/client.ts
rpastro: No. The state warning is happening with our own server, so it could be an issue with our server.  I'm debugging it now to see what is happening.  If it turns out to be an issue on the client I'll open a separate issue.

enisdenjo: But, is it happening on this server? If not, I would say the issue is not related. The only difference in the client with this PR is that it has -1 send, I guess these CLOSED and CLOSING warnings would have been noticed before - or caught by tests.

Still, if it indeed turns out to be a client issue, please open an additional issue then.

rpastro: I found the issue and opened https://github.com/enisdenjo/graphql-ws/issues/66

enisdenjo (approved):

PR #67, opened 2020-11-12
fix(client): One cleanup per subscription

### LINKED ISSUE #66: Socket is closed even though there are pending subscriptions
I'm using graphql-ws together with Apollo client. 

I noticed that for queries and mutations the cancellerRef.current is being invoked twice. Once when the Complete message is received and again from the cleanup function.

As a result state.locks gets decremented twice for the same operation, which causes the socket to be closed even though there are pending operations.

rpastro:
Fixes #66

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -249,6 +249,7 @@ export function createClient(options: ClientOptions): Client {
                 }
 
                 cancellerRef.current = () => {
+                  cancellerRef.current = null;
                   cleanup?.();
                   state.locks--;
                   if (!state.locks) {
@@ -375,6 +376,7 @@ export function createClient(options: ClientOptions): Client {
           }
 
           cancellerRef.current = () => {
+            cancellerRef.current = null;
             cleanup?.();
             state.locks--;
             if (!state.locks) {
--- a/src/client.ts
+++ b/src/client.ts
@@ -243,6 +243,7 @@ export function createClient(options: ClientOptions): Client {
 
                 state.socket.addEventListener('close', listener);
                 function listener(event: CloseEvent) {
+                  cancellerRef.current = null;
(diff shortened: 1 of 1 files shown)


enisdenjo (changes requested):
Great! I like.

I'd also remove the `finally` method from the subscribe promise, this one: https://github.com/enisdenjo/graphql-ws/blob/3f4f836c395139617a268082131084c4f992ba5f/src/client.ts#L543

and add the `cancellerRef.current = null;` in the `throwOnCloseOrWaitForCancel` close listener.

This would fit nicely because there are 2 possible exits for the subscribe promise: it cancels and resolves or the socket closes and it rejects. In both cases, there is nothing to cancel afterwards.

rpastro:
@enisdenjo  I pushed a new patch with your suggestion. Hopefully I got them right. ☺️

src/client.ts
```diff
@@ -497,6 +499,7 @@ export function createClient(options: ClientOptions): Client {
             // either the canceller will be called and the promise resolved
             // or the socket closed and the promise rejected
             await throwOnCloseOrWaitForCancel(() => {
+              cancellerRef.current = null;
```
enisdenjo: Uhm, at this point the canceller is already gone 😄 . This is what I had in mind with https://github.com/enisdenjo/graphql-ws/pull/67#pullrequestreview-529489737: https://github.com/enisdenjo/graphql-ws/pull/67/commits/fe3c6f64b834287174e2b0d45b4429c13ef458e9.

enisdenjo (approved):

src/client.ts
rpastro: That definitely makes more sense.  😄

PR #82, opened 2020-12-03
Update _server_.context.md

slaviczavik:
It's just a draw, better description would be better and I also don't know how to describe types of socket and request fields. Anyway it can be useful for someone who is looking for info how to access a request object.

### DIFF
--- a/docs/interfaces/_server_.context.md
+++ b/docs/interfaces/_server_.context.md
@@ -62,6 +62,15 @@ ___
 An extra field where you can store your own context values
 to pass between callbacks.
 
+An object containing socket and request fields passed from `ws` connection event.
+
+#### Fields:
+
+Name     | Type      |
+-------- | --------- |
+`socket` | WebSocket |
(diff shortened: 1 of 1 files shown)


enisdenjo:
Thanks for the PR; however, I'd like to point out a few things:
- The docs are auto-generated. You should update the relevant comment in code and then run `yarn gendocs` to update the documentation.
- Those extras are not a part of the core server controller. The two fields are actually coming from the, opinionated, convention-based, `ws` implementation in `graphql-ws/lib/use/ws`. The `extra` prop in the context is for you to manipulate.
- [The context extras when using `lib/use/ws` is actually already documented.](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/_use_ws_.extra.md)

I'd say this PR is kinda' unnecessary, would you agree?

slaviczavik:
Absolutely agree, didn't "see" extras documentation before. Closing PR.

Issue #83, opened 2020-12-06
Interoperability suggestions


morris:
Hi!

First of all thanks a lot for taking over this and moving GraphQL over web sockets forward, I believe it's quite important to get some consistency here 🎉 !

I was playing around with this a bit and built an alternate server implementation of the protocol, with the goal of supporting @apollo/client out of the box, with no changes on the client side. I found that there's only a few changes that would allow the latest @apollo/client:

- Allow both `graphql-transport-ws` and `graphql-ws` sub-protocols (the latter is used with @apollo/client)
- Allow `type: "start"` in addition to `type: "subscribe"` for `Subscribe` messages
- Allow `type: "stop"` in addition to `type: "complete"` for client-to-server `Complete` messages
- Use `type: "data"` instead of `type: "next"` for `Next` messages if `socket.protocol === "graphql-ws"`
- Allow the client to send `Subscribe` messages before receiving `ConnectionAck`. This was the "hardest" one, as the server needs to defer handling the received `Subscribe` messages until it actually acknowledges the connection. I resolved this through introducing a promise for the acknowledgment, which the `Subscribe` handlers await before proceeding.

I believe some degree of legacy protocol handling could drive adoption a lot. If we had a server-side GraphQL-over-WS implementation that caters to all popular clients right now, that would be really cool (and graphql-ws could be just that!).

This does not necessarily have to be reflected in the spec and may just go to the implementation. Still, I could imagine adding something like:

- Servers MAY accept `Subscribe` messages before a client received the `ConnectionAck` message in order to support legacy clients.
- Servers MUST NOT handle `Subscribe` messages before sending the `ConnectionAck` message.
- etc.

I did write an alternate implementation from scratch (and this was quite easy thanks to your work on the protocol) but I could easily provide a PR for the above if you're interested. Let know what you think :)

All the best
Morris

enisdenjo:
Given the fact that the client presents supported sub-protocols on the initial WebSocket handshake - the server can choose where to delegate the request (old vs new protocol).

Keeping this in mind, I dont plan on "polluting" a fresh protocol with superfluous additions of an outdated one (same for the implementation).

Sorry if this is not the answer you expected... I am fully aware of the transitioning phase, backwards compatibility and the, somewhat, rocky road to adaptation - but changing a working system "for convenience only" is not what I intend to do.

[...]

PR #92, opened 2021-01-06
feat: log warning in dev when using incorrect protocol

vegetabill:
I was integrating graphql-ws for a hackathon at work and it took me ages to figure out that the problem was I was trying to test with [Altair](https://chrome.google.com/webstore/detail/altair-graphql-client/flnheeellpciglgpaodhkhmapeljopja), a nice GraphiQL-like extension that looked like it supported extensions, but uses the `graphql-ws` protocol. The fact that it fails somewhat silently added time. I eventually found #83 and understood why it's not supported.

Adding a warning in dev mode to help folks like me figure out this faster since the ecosystem is a bit messy for a new person.

### Example Screenshot

Thanks for considering it.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -26,7 +26,12 @@ import {
   ErrorMessage,
   CompleteMessage,
 } from './message';
-import { isObject, isAsyncIterable, areGraphQLErrors } from './utils';
+import {
+  isObject,
+  isAsyncIterable,
+  areGraphQLErrors,
+  logDeveloperWarning,
+} from './utils';
 import { ID } from './types';
 
 export type OperationResult =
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -398,6 +403,11 @@ export function makeServer<E = unknown>(options: ServerOptions<E>): Server<E> {
     opened(socket, extra) {
       if (socket.protocol !== GRAPHQL_TRANSPORT_WS_PROTOCOL) {
         socket.close(1002, 'Protocol Error');
+        logDeveloperWarning(
+          `Invalid protocol: '${socket.protocol}'. 
+          Only ${GRAPHQL_TRANSPORT_WS_PROTOCOL} is supported.
+          For more details see https://github.com/enisdenjo/graphql-ws/issues/83`,
+        );
```
enisdenjo, line 410: No need to have this `logDeveloperWarning`. I try avoiding any logs and rely on exposing as much as possible for the users. Simply put the `if != PROD console.warn` right here.

Also, the indentation in backticks will be preserved. Consider inlining the message.

enisdenjo (changes requested):
One thing that I'd recommend in addition is not even allowing the WS upgrade in case the subprotocol is wrong. You can use the `Sec-WebSocket-Protocol` header to validate the request. [This recipe might help you out.](https://github.com/enisdenjo/graphql-ws#ws-backwards-compat)

Nevertheless, since the lib checks it too - it makes sense to have this helpful warning.

vegetabill:
Awesome thanks!

Issue #102, opened 2021-01-30
socket.removeEventListener('close',..) not present in 4.1.2, (maybe) reason for MaxListenersExceededWarning from Node


philipsoeberg:
**Expected Behaviour**
flawless loop query from this most excellent gql websocket library.

**Actual Behaviour**
Using v4.1.2 iteration 8 of a loop query cause 
`(node:20284) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [WebSocket]. Use emitter.setMaxListeners() to increase limit` in node v12.19

**Debug Information**
It seems change https://github.com/enisdenjo/graphql-ws/blame/c4a3c2326192e6a01db08fba49dc5abe6b119c55/src/client.ts#L372 removed the `socket.removeEventListener('close',..)` previously there, which (maybe?) cause MaxListenersExceededWarning() to emit after using the `Client usage with Promise` Recipe 8+ times. The `MaxListenersExceededWarning` is not emitted in graphql-ws v4.1.1.

**Further Information**
I wrote a simple loop query using the "Client usage with Promise" recipe doing a
```
  for (const x of Array(15).keys()) {
    const UserLoop = await execute(
      {
        operationName: "user",
        query: '{ user { id } }',
        variables: {},
      });
      console.log("UserLoop",x,"id is", UserLoop?.data?.user?.id)
  }
```

Using graphql-ws v4.1.1 everything runs smoothly.
Using graphql-ws v4.1.2 MaxListenersExceededWarning: is emitted after 8th' loop.

Perhaps I am doing something wrong?

enisdenjo:
Hey hey, thanks for reporting this issue! Nope, I am indeed subscribing to the same event on each subscription which causes Node to raise this warning. As you may noticed I am already attempting to fix this with the open draft PR, bear with me! 😄

enisdenjo:
Just a short update, the referenced PR solves the `'close'` listener warnings. However, we do create `'message'` listeners on subscribe too - I guess they would raise the same warning.

So, as soon as I get rid of the `'message'` listener - I'll merge the PR and effectively get rid of all potential `MaxListenersExceededWarning` warnings.

P.S. The reason why `v4.1.1` raised no warnings for you, even though the `'message'` listeners stayed exactly the same, is because your tests showcase single-result queries. As soon as you run the same example, but with streaming operations, a second warning would be raised pointing to the `'message'` listeners.

PR #104, opened 2021-01-31
fix(client): Reduce WebSocket event listeners and add new `message` event for incoming messages

enisdenjo:
Closes: #102

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -21,9 +21,15 @@ export * from './protocol';
 
 export type EventConnecting = 'connecting';
 export type EventConnected = 'connected'; // connected = socket opened + acknowledged
+export type EventMessage = 'message';
 export type EventClosed = 'closed';
 export type EventError = 'error';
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -282,62 +282,73 @@ export function createClient(options: ClientOptions): Client {
...
+                    type: MessageType.ConnectionInit,
+                    payload:
+                      typeof connectionParams === 'function'
+                        ? await connectionParams()
+                        : connectionParams,
+                  }),
+                );
+              } catch (err) {
+                socket.close(
+                  4400,
```
AugustoCalaca, line 319: some reference where I can find out the meaning of the codes sent on socket close?

enisdenjo: The 4XXX close event codes are reserved for applications. For convenience, I use HTTP status codes prefixed with a 4.

4400 can be seen as a HTTP `400 Bad Request`.

PR #110, opened 2021-02-12
fix: Define entry points through the `exports`  field and use `.mjs` suffixed ESM imports

n1ru4l:
For ESM to correctly function in a Node.js environment the import statements must end with `.mjs`. Also, all non-top level exports must be declared within the `package.json` `exports` map.

Prior to this change trying to run the following:

**main.mjs**
```js
import { execute, subscribe } from 'graphql';
import https from 'http';
import ws from 'ws';
import { useServer } from 'graphql-ws/lib/use/ws';

```

```
node main.mjs
```

resulted in the following error:

```
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/laurin/projects/graphql-ws-test/node_modules/graphql-ws/lib/use/ws' imported from /Users/laurin/projects/graphql-ws-test/main.mjs
Did you mean to import graphql-ws-test/node_modules/graphql-ws/lib/use/ws.js?
    at finalizeResolution (internal/modules/esm/resolve.js:276:11)
    at moduleResolve (internal/modules/esm/resolve.js:699:10)
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:810:11)
    at Loader.resolve (internal/modules/esm/loader.js:85:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:229:28)
    at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:51:40)
    at link (internal/modules/esm/module_job.js:50:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```


package.json
```diff
@@ -71,9 +81,11 @@
     "eslint": "^7.19.0",
     "eslint-config-prettier": "^7.2.0",
     "eslint-plugin-prettier": "^3.3.1",
+    "glob": "7.1.6",
     "graphql": "^15.5.0",
     "jest": "^26.6.3",
     "prettier": "^2.2.1",
+    "replacestream": "4.0.3",
```
enisdenjo, line 88: Please don't install deps with exact versions.

scripts/esm-post-process.js
```diff
@@ -0,0 +1,20 @@
...
+    parts.pop();
+    const mjsFileName = parts.join(".") + ".mjs"
+    fs.createReadStream(path.join(__dirname, "..", filePath))
+      .pipe(replaceStream(/from '(\.?\.\/[^']*)'/g, "from '$1.mjs'"))
+      .pipe(fs.createWriteStream(path.join(__dirname, "..", mjsFileName)))
+      .on("close", () => {
+        fs.unlinkSync(filePath);
+      });
+  }
+});
```
enisdenjo, line 20: Wondering if we can pimp our Rollup or TS build configurations instead of doing custom post-processing?

enisdenjo (changes requested):
Awesome, thanks for pushing this PR! I havent really tested ESM support. 😅

package.json
enisdenjo, line 88: Please dont install deps with exact versions.

enisdenjo (changes requested):

scripts/esm-post-process.js
n1ru4l: I dug a bit, but it seems like this is not possible with typescript at the moment: https://github.com/microsoft/TypeScript/issues/18442

Someone posted a rollup solution over here: https://github.com/microsoft/TypeScript/issues/18442#issuecomment-749896695

However, afaik rollup is rather used for building a single file instead of transpiling many files? But maybe I am wrong?

The script seemed the easiest solution with the fewest additional dependencies to me (that also works on all os platforms).

package.json
n1ru4l: A sorry! I tend to pin them and let a bot update the pinned version for dev dependencies. WIll fix

enisdenjo: I want to add the Dependabot too; however, [Yarn 2 support is still in works](https://github.com/dependabot/dependabot-core/issues/1297)...

n1ru4l: You should try to use renovate, I prefer it over dependabot. haven't used it with yarn2 though as I don't like it :D

scripts/esm-post-process.js
enisdenjo: Yeah, TS is still struggling and, you're right, Rollup is for bundling to a single file.

The script is nice, very straight forward. Was just a thought.

n1ru4l: My only concern would be too many open files at the same time, however, I guess given the number of files we got right now that shouldn't be too bad. It can always be optimized later.

enisdenjo: Thats fine for now. Premature optimization is the root of all evil haha.

We can get correct ES modules with `.mjs` imports using this Rollup config:
```js
import typescript from '@rollup/plugin-typescript';
export default {
  preserveModules: true,
  input: './src/index.ts',
  output: [
    {
      dir: 'lib',
      format: 'es',
      entryFileNames: '[name].mjs',
    },
  ],
  plugins: [typescript()],
};
```

**However,** [Rollup seems to "force" the use of `tslib`](https://github.com/rollup/plugins/issues/265#issuecomment-599339539). Because of this, it bundles a `node_modules` folder with `tslib` which I dont like at all. Firstly because we need the `__asyncValues` helper function [exclusively at one place](https://github.com/enisdenjo/graphql-ws/blob/9daef127ad1a1205abba9f1e896921492552d39b/src/server.ts#L675), and secondly, it looks ugly.

enisdenjo (approved):
Looks good to me! Thank you @n1ru4l.

PR #128, opened 2021-02-25
docs: use return early to improve the readability of the example

AugustoCalaca:


enisdenjo (approved):
I like! Thanks @AugustoCalaca.

PR #143, opened 2021-03-25
feat(client): `url` option accepts a function or a Promise

### LINKED ISSUE #146: Supporting url in ClientOptions to be a function that returns a Promise
### Story

Currently we have a use case that there is service that's checking csrf token upon the initial request, and it's expecting the csrf token to be part of the query string (url in this context). Basically we are trying to make the url follow this pattern wss://graphql-ws?csrf=<token>.

The csrf token could expire, so when there's an invalid token error, we would like to refetch and recreate a graphql-ws client with the updated token. Therefore, to make it recreate the client using the updated token, we would like to make url to support a function type that returns a Promise, so it can be dynamically updatable.

Specifically, we'd like our use case to be like this:
`createClient({ url: () => new Promise((resolve) => { resolve(getUrlWithLatestCsrfToken()) }), lazy: true, })`

### Acceptance criteria

- `url` in `ClientOptions` can be a function that returns a Promise

edwardfeng-db:
Closes #145, closes #146 

Updated `url` in `ClientOptions` to support function type that returns a Promise, the same way as `connectionParams`. The use case is that it can be used together with automatic reconnect on abnormal socket closure such that the url can be dynamically changing each time it reconnects.

Our specific use case: we have a service that checks the csrf token as part of the query string, so we have to add the csrf token to the `url`. The token might expire, and in that case, we would like to take advantage of the automatic reconnect functionality to be able to dynamically pick up an updated csrf token, and this can be achieved with a function that returns a Promise.

Specifically, we'd like to use it as:
`createClient({
  url: () => new Promise((resolve) => { resolve(getUrlWithLatestCsrfToken()) }),
  lazy: true,
})`

### DIFF
--- a/docs/interfaces/client.clientoptions.md
+++ b/docs/interfaces/client.clientoptions.md
@@ -213,9 +213,12 @@ ___
 
 ### url
 
-• **url**: *string*
+• **url**: *string* \| (() => *Promise*<*string*> | <*string*>)
 
 URL of the GraphQL over WebSocket Protocol compliant server to connect.
+
+The case of returning a Promise can be used together with automatic reconnect upon abnormal
+socket closure if you need your url to be dynamically changing each time it reconnects. 
 
 ___
 
--- a/docs/interfaces/client.clientoptions.md
+++ b/docs/interfaces/client.clientoptions.md
@@ -213,12 +213,17 @@ ___
 
 ### url
 
-• **url**: *string* \| (() => *Promise*<*string*> | <*string*>)
+• **url**: *string* \| () => *string* \| *Promise*<string\>
 
 URL of the GraphQL over WebSocket Protocol compliant server to connect.
 
-The case of returning a Promise can be used together with automatic reconnect upon abnormal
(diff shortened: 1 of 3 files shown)


enisdenjo:
Heads up! I've edited your description and linked the related issues.

Please refer to this article for more information:
https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

edwardfeng-db:
Oh thanks @enisdenjo! Sorry about this, I thought that `Closes #` is only for when you merge the PR and close the issue. My bad. Thanks for the information.

enisdenjo:
It is! But you put the `Closes #` right away so that you link the issue with the PR. If the PR gets merged, issue is closed without further action; and if the PR gets declined, the issue stays open but still has a reference that there was a failed attempt to fix it.

enisdenjo (approved):
Thanks for the PR! I only made minimal adjustments simply because of my OCD.

PR #150, opened 2021-03-26
perf(client): Focus subscription message listeners on `id`

michelepra:
If client is used for a lot of simultaneous subscription, ther's unnecessary overload due to check MessageType and id for every subscription handler


enisdenjo (approved):
Nice one, thanks @michelepra, this looks great!

PR #171, opened 2021-04-26
feat(server): allow custom message stringification and parsing

airhorns:
If you want to use a different transport than JSON over the websocket, or use an extension of JSON to support things like bigints, it's handy to be able to control the stringification and unstringifiation of the messages going across this socket! This adds options to the server to allow overriding the `stringifyMessage` and `parseMessage` functions the server uses so userland can hook in.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -24,8 +24,8 @@ import { GRAPHQL_TRANSPORT_WS_PROTOCOL } from './protocol';
 import {
   Message,
   MessageType,
-  stringifyMessage,
-  parseMessage,
+  stringifyMessage as defaultStringifyMessage,
+  parseMessage as defaultParseMessage,
   SubscribeMessage,
   NextMessage,
   ErrorMessage,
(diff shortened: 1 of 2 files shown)


enisdenjo (changes requested):
Cool stuff; but, [the Protocol requires the messages to be stringified JSONs](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#communication). Therefore, I wouldnt feel safe allowing full control over the stringification process for 2 reasons: not parsing/stringifying JSONs and not doing validation.

I'd suggest supporting a custom [JSON.parse `reviver` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#using_the_reviver_parameter) and the [JSON.stringify `replacer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter) counterpart instead.

Oh and, I guess the client needs the same possibilities?

airhorns:
Ah ha -- that makes sense! I will switch to that approach, my specific use case is support for bigints and/or big decimals so that would work fine. Thanks for the feedback!

PR #172, opened 2021-04-27
feat: Support custom JSON message `reviver` and `replacer`

airhorns:
If you want to use a fancier types inside your JSON over the websocket like bigints or dates, it's handy to be able to control the stringification and unstringifiation of the messages going across this socket! This adds options to the server to allow overriding the `revive` and `replace` functions the server uses in JSON.stringify and JSON.parse so userland can hook in and alter the behaviour. We stay compatible with the graphql-ws protocol by sticking with JSON but still allow flexibility to marshal fancier types to and from JSON easily.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -14,6 +14,8 @@ import {
   parseMessage,
   stringifyMessage,
   SubscribePayload,
+  JSONMessageReviver,
+  JSONMessageReplacer,
 } from './common';
 import { isObject } from './utils';
 
@@ -236,6 +238,18 @@ export interface ClientOptions {
    * Reference: https://gist.github.com/jed/982883
    */
   generateID?: () => ID;
+  /**
+   * An optional override for the JSON.parse function used to hydrate
+   * incoming messages to this client. Useful for parsing custom datatypes
+   * out of the incoming JSON.
+   */
+  jsonMessageReviver?: JSONMessageReviver;
+  /**
+   * An optional override for the JSON.stringify function used to serialize
+   * outgoing messages from this client. Useful for serializing custom
+   * datatypes out to the client.
+   */
+  jsonMessageReplacer?: JSONMessageReplacer;
 }
 
 /** @category Client */
@@ -298,6 +312,8 @@ export function createClient(options: ClientOptions): Client {
         return v.toString(16);
       });
     },
+    jsonMessageReplacer: replacer,
+    jsonMessageReviver: reviver,
   } = options;
 
   let ws;
@@ -412,13 +428,16 @@ export function createClient(options: ClientOptions): Client {
           socket.onopen = async () => {
             try {
               socket.send(
-                stringifyMessage<MessageType.ConnectionInit>({
-                  type: MessageType.ConnectionInit,
-                  payload:
-                    typeof connectionParams === 'function'
-                      ? await connectionParams()
-                      : connectionParams,
-                }),
+                stringifyMessage<MessageType.ConnectionInit>(
+                  {
+                    type: MessageType.ConnectionInit,
+                    payload:
+                      typeof connectionParams === 'function'
+                        ? await connectionParams()
+                        : connectionParams,
+                  },
+                  replacer,
+                ),
               );
             } catch (err) {
               socket.close(
@@ -431,7 +450,7 @@ export function createClient(options: ClientOptions): Client {
           let acknowledged = false;
           socket.onmessage = ({ data }) => {
             try {
-              const message = parseMessage(data);
+              const message = parseMessage(data, reviver);
               emitter.emit('message', message);
               if (acknowledged) return; // already connected and acknowledged
 
@@ -600,21 +619,27 @@ export function createClient(options: ClientOptions): Client {
             });
 
             socket.send(
-              stringifyMessage<MessageType.Subscribe>({
-                id,
-                type: MessageType.Subscribe,
-                payload,
-              }),
+              stringifyMessage<MessageType.Subscribe>(
+                {
+                  id,
+                  type: MessageType.Subscribe,
+                  payload,
+                },
+                replacer,
+              ),
             );
 
             releaser = () => {
               if (!done && socket.readyState === WebSocketImpl.OPEN)
                 // if not completed already and socket is open, send complete message to server on release
                 socket.send(
-                  stringifyMessage<MessageType.Complete>({
-                    id,
-                    type: MessageType.Complete,
-                  }),
+                  stringifyMessage<MessageType.Complete>(
+                    {
+                      id,
+                      type: MessageType.Complete,
+                    },
+                    replacer,
+                  ),
                 );
               locks--;
               done = true;
--- a/src/common.ts
+++ b/src/common.ts
@@ -191,23 +191,48 @@ export function isMessage(val: unknown): val is Message {
 }
 
 /**
+ * Function for transforming values within a message during JSON parsing
+ * The values are produced by parsing the incoming raw JSON.
+ *
+ * Read more about using it:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#using_the_reviver_parameter
+ *
+ * @category Common
+ */
+export type JSONMessageReviver = (this: any, key: string, value: any) => any; // eslint-disable-line @typescript-eslint/no-explicit-any
+
+/**
  * Parses the raw websocket message data to a valid message.
  *
  * @category Common
  */
-export function parseMessage(data: unknown): Message {
+export function parseMessage(
+  data: unknown,
+  reviver?: JSONMessageReviver,
+): Message {
   if (isMessage(data)) {
     return data;
   }
   if (typeof data !== 'string') {
     throw new Error('Message not parsable');
   }
-  const message = JSON.parse(data);
+  const message = JSON.parse(data, reviver);
   if (!isMessage(message)) {
     throw new Error('Invalid message');
   }
   return message;
 }
+
+/**
+ * Function that allows customization of the produced JSON string
+ * for the elements of an outgoing `Message` object.
+ *
+ * Read more about using it:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter
+ *
+ * @category Common
+ */
+export type JSONMessageReplacer = (this: any, key: string, value: any) => any; // eslint-disable-line @typescript-eslint/no-explicit-any
 
 /**
  * Stringifies a valid message ready to be sent through the socket.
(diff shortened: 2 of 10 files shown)


src/common.ts
```diff
@@ -190,35 +190,48 @@ export function isMessage(val: unknown): val is Message {
...
   }
-  const message = JSON.parse(data);
+  const message = JSON.parse(data, reviver);
   if (!isMessage(message)) {
     throw new Error('Invalid message');
   }
   return message;
 }
 
+export type JSONReplacer = (
```
enisdenjo, line 218: I feel like `JSONMessageReplacer` is a better and more explicit name. You're altering the stringification process of exactly only transported messages. Also, favouring documentation, a description would be nice.

```suggestion

/**
 * Function that alters the behavior of the outgoing `Message`
 * stringification process.
 *
 * Read more about using it:
 * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter
 *
 * @category Common
 */
export type JSONMessageReplacer = (
```

src/client.ts
```diff
@@ -236,6 +238,18 @@ export interface ClientOptions {
...
+   * incoming messages to this client. Useful for parsing custom datatypes
+   * out of the incoming JSON.
+   */
+  reviver?: JSONReviver;
+  /**
+   * An optional override for the JSON.stringify function used to serialize
+   * outgoing messages from this client. Useful for serializing custom
+   * datatypes out to the client.
+   */
+  replacer?: JSONReplacer;
```
enisdenjo, line 252: I prefer longer, but more descriptive names. Especially for options which can grow:
```suggestion
  jsonMessageReviver?: JSONMessageReviver;
  /**
   * An optional override for the JSON.stringify function used to serialize
   * outgoing messages from this client. Useful for serializing custom
   * datatypes out to the client.
   */
  jsonMessageReplacer?: JSONMessageReplacer;
```

src/server.ts
```diff
@@ -364,6 +366,18 @@ export interface ServerOptions<E = unknown> {
...
+   * incoming messages to this server. Useful for parsing custom datatypes
+   * out of the incoming JSON.
+   */
+  reviver?: JSONReviver;
+  /**
+   * An optional override for the JSON.stringify function used to serialize
+   * outgoing messages to from server. Useful for serializing custom
+   * datatypes out to the client.
+   */
+  replacer?: JSONReplacer;
```
enisdenjo, line 380: Same as with the client-side; I prefer longer, but more descriptive names. Especially for options which can grow:
```suggestion
  jsonMessageReviver?: JSONMessageReviver;
  /**
   * An optional override for the JSON.stringify function used to serialize
   * outgoing messages to from server. Useful for serializing custom
   * datatypes out to the client.
   */
  jsonMessageReplacer?: JSONMessageReplacer;
```

src/tests/client.ts
```diff
@@ -329,6 +329,51 @@ it('should not send the complete message if the socket is not open', async () =>
...
+    onNonLazyError: noop,
+    replacer: (_, value) => {
+      if (typeof value == 'string') {
+        return value.toUpperCase();
+      }
+      return value;
+    },
+    reviver: (_, value) => {
+      if (typeof value == 'string') {
+        return value.toLowerCase();
```
enisdenjo, line 360: [The GraphQLSchema `value` field returns the lowercase `"value"`.](https://github.com/gadget-inc/graphql-ws/blob/73bed5e780f1f14cbb117cdb63091f66d77755c8/src/tests/fixtures/simple.ts#L31) This test has no effect on the result, consider maybe using `toUpperCase()` to visualise the alteration.

src/tests/client.ts
```diff
@@ -329,6 +329,51 @@ it('should not send the complete message if the socket is not open', async () =>
   await sub.waitForComplete();
 });
 
+it('should use a custom json reviver and replacer function', async () => {
```
enisdenjo, line 332: You are actually not testing the `replacer` function, only the `reviver`. You can do so by using the `server.waitForMessage` and expecting the replacer results there.

src/tests/client.ts
```diff
@@ -329,6 +329,51 @@ it('should not send the complete message if the socket is not open', async () =>
...
+        return value.toUpperCase();
+      }
+      return value;
+    },
+    reviver: (_, value) => {
+      if (typeof value == 'string') {
+        return value.toLowerCase();
+      }
+      return value;
+    },
```
enisdenjo, line 345: Consider dropping the `startTServer` `replacer` and `reviver` options for consistency and a client focused test.
```suggestion
```

src/tests/server.ts
```diff
@@ -1527,6 +1527,56 @@ describe('Subscribe', () => {
...
+    );
+
+    await client.waitForMessage(({ data }) => {
+      expect(data).toMatchInlineSnapshot(`"{\\"type\\":\\"CONNECTION_ACK\\"}"`);
+    });
+
+    client.ws.close(4321, 'Byebye');
+  });
+
+  it('should use a custom JSON reviver function', async () => {
```
enisdenjo, line 1556:
```suggestion
  it('should use a custom JSON message reviver function', async () => {
```

src/tests/server.ts
```diff
@@ -1527,6 +1527,56 @@ describe('Subscribe', () => {
...
+
+    const client = await createTClient(url);
+    client.ws.send(
+      stringifyMessage<MessageType.ConnectionInit>({
+        type: MessageType.ConnectionInit,
+      }),
+    );
+
+    await client.waitForMessage(({ data }) => {
+      expect(data).toMatchInlineSnapshot(`"{\\"type\\":\\"CONNECTION_ACK\\"}"`);
```
enisdenjo, line 1550: Simpler?
```suggestion
      expect(data).toBe('"{"type":"CONNECTION_ACK"}"');
```

src/tests/server.ts
```diff
@@ -1527,6 +1527,56 @@ describe('Subscribe', () => {
 
     await server.waitForComplete();
   });
+
+  it('should use a custom JSON replacer function', async () => {
```
enisdenjo, line 1531:
```suggestion
  it('should use a custom JSON message replacer function', async () => {
```

enisdenjo (changes requested):
Almost there, a few remarks and this is ready to be merged.

Also, please run `yarn gendocs` to reflect the documentation updates.

src/tests/client.ts
airhorns: Ah my mistake, I thought it would be implicitly tested by transforming the keys of the JSON messages like `type` => `TYPE`. Just tested though and you are right, the replacer only replaces values:

```
> const uppercaser = (_, value) => {
...       if (typeof value == 'string') {
.....         return value.toUpperCase();
.....       }
...       return value;
...     }
undefined
> JSON.stringify({foo: "bar"}, uppercaser)
'{"foo":"BAR"}'
```

Will fix

src/tests/server.ts
airhorns: I used an inline snapshot because I always find getting the quotes and escapes right is hard and I'd rather just `yarn test -u` and let jest write exactly the string that will work. Seems easier to maintain but happy to fix it if you feel strongly!

enisdenjo: Getting which quotes and escapes right? With the `toBe` there is nothing to escape.

Furthermore, the snapshot here will change only if the test case changes which means you'll have to change the test anyway. I feel you would have a harder time decoding the escapes when looking at the tests, comparing the snapshots, having to do `-u`; in comparison to a plain JSON literal.

If the JSON was large, then it'd be fine; but with these 25 characters, I'd like to keep it inline please.

src/tests/client.ts
enisdenjo: I prefer tests being isolated to the problem or scope. For example, if only the server starts failing, due to this integration, the client tests will fail in addition to the server's (2 failed tests, 1 scope). You wouldnt know whats really failing until you poke around. However, if the tests were kept isolated, the client will fail independently of the server and vice versa (1 failed test, 1 scope; 2 failed tests, 2 scopes). Having said that, please test the client here exclusively. (I dont mind the single test for both `reviver` and `replacer`, just dont mix the server in)

You can always transform a different message to avoid validation issues (like the `query { value }` when `MessageType = subscription`). Yeah, might be a few extra lines of code, but we keep the tests scoped.

enisdenjo: Oh and,
> That said, these options are still necessary on the server because the client will uppercase the connection init and downcase the connection ack message as well, so the server needs to properly transform those in order to establish the connection that then lets the actual test message work.

Keeping it simple, you can just separate the `replacer` test, check the message with `server.waitForMessage`, and let it fail due to validation. Still focused, and no harm done.

src/tests/server.ts
enisdenjo: Additionally to getting rid of the escapes, you also drop the extra quotes. So the suggestion is actually:
```suggestion
      expect(data).toBe('{"type":"CONNECTION_ACK"}');
```

src/tests/client.ts
airhorns: I can't find an implementation of `server.waitForMessage` -- am I blind or are you maybe thinking of `client.waitForMessage`? I tried writing my own `server.waitForMessage`, but it's tricky because the first thing the server does when it goes to handle a message is try to parse it, which validates it. If there's no server side reviver that undoes whatever the client does to the message in the test then this validation fails and no event ever gets emitted for the test to pick up, error or otherwise. I haven't come up with a good replacer that makes a noticeable difference in the JSON that the test can assert on that doesn't also break the expectations of the library if the changes aren't undone before validation. 

That's why I wrote the less-than-ideally-isolated test, it just seems a bit easier to let the system work as intended to test this otherwise not super important feature along the happy path, since really it's the JSON builtin that is doing the actual work and we just need to test that the message was transformed at all.

src/tests/server.ts
airhorns: To each their own, changed!

enisdenjo:
Hey @airhorns, since I had the tests and utilities in my head, I refactored them and cleaned up a bit to expedite this PR for you. However, I dont have push rights for the `gadget-inc/graphql-ws` repo and therefore cannot include my changes here. 

Could it be possible for you to give me push rights so that I can add my commits and merge this PR? 😄

src/tests/client.ts
enisdenjo: My bad! Had `startTServer` in my head. 😅

Please see https://github.com/enisdenjo/graphql-ws/pull/172#issuecomment-829520428.

airhorns:
Yessir! Feel free to just close this PR or amend the commit on another branch no worries!

enisdenjo (approved):

enisdenjo:
Brilliant, thanks! In case you're interested, check how I isolated and simplified the client tests here: https://github.com/enisdenjo/graphql-ws/pull/172/commits/73c6dade0645e0a476aa47dcec17a98e3950b8da.

Many thanks for your great additions, @airhorns! 🚀

PR #179, opened 2021-05-08
feat: Support custom file uploaders with `sendMessage(socket, message, files)`

D1plo1d:
This PR adds a `sendMessage(socket, message, files)` option to the Client to allow developers to explore uploading files over websockets either in-band through a wrapping protocol or out-of-band (eg. through [S3 PreSigned Uploads](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html)).

The `#__graphql_file__:${id}` file pointers are compatible with [Apollo's Upload Scalar](https://www.apollographql.com/docs/apollo-server/data/file-uploads/).

Initial server support for websocket file upload will be provided in Rust's Async GraphQL next release ([async-graphql PR here](https://github.com/async-graphql/async-graphql/pull/501)).

It is not expected that this will be the final revision of file uploads in GraphQL Websockets. Instead this PR is intended to open the door to further experimentation to explore the solution space of websocket file uploads with what is hoped to be a minimal enough addition to graphql-ws that it does not warrant code bloat concerns.

#### Example Usage

```ts
const client = createClient({
  url: 'wss://example.com',
  sendMessage: (socket, message, files) => {
    if (files.length > 0) uploadFileToS3(files)

    socket.send(message)
  }
})
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -90,6 +90,13 @@ export type EventClosedListener = (event: unknown) => void;
  */
 export type EventErrorListener = (error: unknown) => void;
 
+export type FileLike = File | Blob | FileList;
+export type SendMessage = (
+  socket: WebSocket,
+  message: string,
+  files: FileLike[],
+) => void;
+
 /** @category Client */
 export type EventListener<E extends Event> = E extends EventConnecting
   ? EventConnectingListener
@@ -250,6 +257,7 @@ export interface ClientOptions {
    * datatypes out to the client.
    */
   jsonMessageReplacer?: JSONMessageReplacer;
+  sendMessage?: SendMessage;
 }
 
 /** @category Client */
@@ -314,6 +322,12 @@ export function createClient(options: ClientOptions): Client {
     },
     jsonMessageReplacer: replacer,
     jsonMessageReviver: reviver,
+    sendMessage = (socket, message, files) => {
+      if (files.length > 0) {
+        throw new Error('Cannot upload files with default sendMessage handler');
+      }
+      socket.send(message);
+    },
   } = options;
 
   let ws;
@@ -618,16 +632,31 @@ export function createClient(options: ClientOptions): Client {
               }
             });
 
-            socket.send(
-              stringifyMessage<MessageType.Subscribe>(
-                {
-                  id,
-                  type: MessageType.Subscribe,
-                  payload,
-                },
-                replacer,
-              ),
+            const files: FileLike[] = [];
+            const uploadReplacer = (key: string, originalValue: any) => {
+              let value = originalValue;
+              if (
+                value instanceof File ||
+                value instanceof Blob ||
+                value instanceof FileList
+              ) {
+                const fileUploadID = `#__graphql_file__:${files.length}`;
+                files.push(value);
+                value = fileUploadID;
+              }
+              return replacer ? replacer(key, value) : value;
+            };
+
+            const message = stringifyMessage<MessageType.Subscribe>(
+              {
+                id,
+                type: MessageType.Subscribe,
+                payload,
+              },
+              uploadReplacer,
             );
+
+            sendMessage(socket, message, files);
 
             releaser = () => {
               if (!done && socket.readyState === WebSocketImpl.OPEN)


D1plo1d:
> However, if the WS connection is expected to really do the upload by streaming the bytes through the socket, then we need Protocol alterations supporting this (since, then, not all messages are JSON strings).

Sorry for the delay. Yeah, so to remove any confusion around my broken S3 example (your right, that should be async) - a file upload protocol is my primary application of this. I'm layering `graphql-ws` over a [modified SaltyRTC protocol](https://github.com/tegapp/teg/blob/37b8af183acdc3c7887ce0bc5f094e6c64e683bc/packages/teg-web-ui/src/webrtc/saltyRTCChunk.ts) to send GraphQL and file uploads over a WebRTC connection.

I don't expect many people to want to do this but I would prefer if it was doable without modification to the `graphql-ws` lib.

> Do a base64 encoding, or send a byte array, through the query variables

This is actually what I was doing previously. Unfortunately, testing found that it created performance issues server-side (I'm using rust's [async-graphql](https://async-graphql.github.io/async-graphql/en/index.html)) because the GraphQL server implementation did not expect 10s to 100s of MBs of file uploads in variables.

It also precluded the optimization of streaming each `Upload` directly to a tempfile. The only way around this second limitation would be to make my SaltyRTC protocol parsing code aware of the GraphQL schema so it could find Uploads via schema inspection but my architecture is much cleaner atm with those two decoupled.

So including the file upload as a String or byte array while workable for small files was not preferred.

> If I understand the file pointer IDs correctly, they should be unique across the file storage and the database should point to that exact file - this is how 2. would work

This is a very interesting idea. I'm still trying to wrap my head around a version of this that would work for my WebRTC usage. One aspect I quite like about the current implementation is that if the GraphQL request errors for any reason then the tempfiles get unlinked/deleted when they go out of scope. So to maintain that I'd want some way to associate a set of files with a stringified JSON upload before sending both in a SaltyRTC message to the server.

I don't have a solution yet for associating the stringified JSON to the files but I figured an update on my work in progress thinking might help. Hope that makes a bunch more sense then my previous post! Thanks for the help!

enisdenjo:
Interesting stuff! I like.

Sending encodings/byte arrays in inefficient, you're right. Was just thinking about ways of not modifying the Protocol yet supporting the upload needs.

Regarding my example (upload then subscribe), the only real efficient way of having this work reliably is to have the `uploadFile from './my-server'` be really dumb. What I mean by this is that the `uploadFile` quite literally only puts the file in a set location on the server and assigns it a unique ID. Doing this, the _only_ thing that could go wrong is the upload itself - making sure that the subscribe always follows.

Now, keeping the `uploadFile` ultra dumb; you may do all your processing in the GQL execution context itself. The only GQL execution prerequisite would be for that file with that unique ID to exist (otherwise fail instantly). This way, you really do all file manipulation as a part of the GQL request and can leverage all its sexiness.

Furthermore, if you need the upload to be done through a socket (WebRTC say), why not simply have another socket reserved for file uploads exclusively? It can implement exactly the subprotocol you need with all rules/chunking/compression techniques for an efficient upload; oh and, `graphql-ws` would, hence, stay untouched.

One thing that I'd like to avoid is doing file uploads with the GraphQL over WebSocket Protocol. Mainly because that adds another, bigger, layer of complexity. By cutting of the "all messages are stringified JSONs" part, we need to specify the alternate for file uploads and quite literally design the **whole** upload process, simply because the file cannot be a part of the JSON. It has to be streamed separately and then connected. Which, ultimately, is exactly what we're trying to do here (upload separately and then connect with a file pointer).

P.S. I am available on [GraphQL Slack](https://slack.graphql.org/) in the `#graphql-ws` channel. If you'd like, we can chat about this to speed stuff up. (Feel free to DM me too)

enisdenjo:
@D1plo1d any news?

D1plo1d:
@enisdenjo Sorry about that - I've been meaning to follow up but I got pulled away by a bunch of things. I'd love to chat with you about this though!

I think things will proverbially be less on fire here by next week. Any days that work better for you?

enisdenjo:
Thats fine. I know the feeling. 😄 

No specific days, I am busy rather always haha, so, just feel free to DM me or write in the channel - I'll join in as soon as I can.

enisdenjo:
Closing this for now. Needs further discussion and proper thought investment. @D1plo1d feel free to open a discussion, or even an issue if you have a concrete example that has to be solved on WebSocket transport level.

Furthermore, my offer to discuss this on Slack still stands, would be happy to talk about this with you!

PR #192, opened 2021-05-31
docs: use port 4000 and drop tls for clarity

rafaelpernil2:
All code snippets in README.md used wss:// as protocol insead of ws://

I find way more useful providing code snippets that work in a test environment without a certificate setup than examples for production environments. With this approach, users can try out the library more easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ import ws from 'ws'; // yarn add ws
 import { useServer } from 'graphql-ws/lib/use/ws';
 
 const server = new ws.Server({
-  port: 80,
+  port: 4000,
   path: '/graphql',
 });
 
@@ -67,7 +67,7 @@ useServer(
   server,
 );
 
-console.log('Listening to port 80');
+console.log('Listening to port 4000');
 ```
 
 ##### With [uWebSockets.js](https://github.com/uNetworking/uWebSockets.js)
@@ -85,9 +85,9 @@ uWS
       { schema, roots },
     ),
   )
-  .listen(80, (listenSocket) => {
+  .listen(4000, (listenSocket) => {
     if (listenSocket) {
-      console.log('Listening to port 80');
+      console.log('Listening to port 4000');
     }
   });
 ```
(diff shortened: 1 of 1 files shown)


enisdenjo (commented):
None of the URLs you changed exist, they are imaginary and only for example purposes. Therefore, I intentionally use the `wss` protocol because I want to incentivise the users to use SSL/TLS for their production endpoints.

rafaelpernil2:
I get your point, but I find easy to miss removing that s when copying one of those code snippets and chaging the endpoint (without wss://). At least that's what happened to me.

Also, the code snippets are not prepared for SSL/TLS, so they wouldn't work as provided. I know you have to change the entire endpoint, but they don't represent production scenarios and those tiny details help when learning.

Another interesting approach would be to change all the examples to ws:// and add other examples for wss using an https server and showing some basic settings...

Thanks for your quick response! :blush:

enisdenjo:
> Another interesting approach would be to change all the examples to ws:// and add other examples for wss using an https server and showing some basic settings...

I'd say that is an overkill, I think you agree with me about that too. 😄 

> Also, the code snippets are not prepared for SSL/TLS, so they wouldn't work as provided.

For all server recipes I use the 443 port and whenever I need the HTTP server I import `from 'https'`. But yeah, you're right about the fact that setting up a HTTPS server needs more configuration than just that.

enisdenjo:
@rafaelpernil2 I think you're right, the docs should be easy to consume and easier to copy over. I'd just ask you one more thing: could you change the ports to something more sensible too (not 443s)? Feel free to chose any number. 😄

rafaelpernil2:
@enisdenjo Sure, I'll change them to 4000 for example. I think it's important to show a different port than 80 (HTTP) or 443 (HTTPS) because it shows how ports are used and affect the code. You know: "I see a 4000 here in the port property so... It must have 4000 in the url"

rafaelpernil2:
I already made the change :)

enisdenjo (approved):

enisdenjo:
Great stuff, thanks!

PR #196, opened 2021-06-01
feat(uWebSockets): Add `persistedRequest` to context extra and deprecate uWS's stack allocated `request`

maxpain:
Due to [this limitation](https://github.com/uNetworking/uWebSockets.js/discussions/84) (for performance optimizations), we can't use [uWS.HttpRequest ](https://unetworking.github.io/uWebSockets.js/generated/interfaces/httprequest.html) out of  scope of `upgrade` callback, and it wasn't possible to get some headers from `upgradeReq`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -769,57 +769,6 @@ wsServer.on('connection', (socket, request) => {
 
 </details>
 
-<details id="uws-auth-handling">
-<summary><a href="#uws-auth-handling">🔗</a> Server usage with <a href="https://github.com/uNetworking/uWebSockets.js">uWebSockets.js</a> and custom auth handling</summary>
-
-```typescript
-import uWS from 'uWebSockets.js'; // yarn add uWebSockets.js@uNetworking/uWebSockets.js#<tag>
-import cookie from 'cookie';
-import { makeBehavior, Request } from 'graphql-ws/lib/use/uWebSockets';
-import { schema } from './my-graphql-schema';
-import { validate } from './my-auth';
-
-// your custom auth
-class Forbidden extends Error {}
-function handleAuth(request: Request) {
-  // do your auth on every subscription connect
-  const good = validate(request.headers['authorization']);
-  // or const { iDontApprove } = session(cookie.parse(request.headers.cookie));
-  if (!good) {
-    // throw a custom error to be handled
-    throw new Forbidden(':(');
-  }
-}
-
-uWS
-  .App()
-  .ws(
-    '/graphql',
-    makeBehavior({
-      schema,
-      onConnect: async (ctx) => {
-        // do your auth on every connect
-        await handleAuth(ctx.extra.request);
-      },
-      onSubscribe: async (ctx) => {
-        // or maybe on every subscribe
-        await handleAuth(ctx.extra.request);
-      },
-      onNext: async (ctx) => {
-        // haha why not on every result emission?
-        await handleAuth(ctx.extra.request);
-      },
-    }),
-  )
-  .listen(4000, (listenSocket) => {
-    if (listenSocket) {
-      console.log('Listening to port 4000');
-    }
-  });
-```
-
-</details>
-
 <details id="express">
 <summary><a href="#express">🔗</a> <a href="https://github.com/websockets/ws">ws</a> server usage with <a href="https://github.com/graphql/express-graphql">Express GraphQL</a></summary>
 
--- a/README.md
+++ b/README.md
@@ -769,6 +769,57 @@ wsServer.on('connection', (socket, request) => {
 
 </details>
 
+<details id="uws-auth-handling">
+<summary><a href="#uws-auth-handling">🔗</a> Server usage with <a href="https://github.com/uNetworking/uWebSockets.js">uWebSockets.js</a> and custom auth handling</summary>
+
+```typescript
+import uWS from 'uWebSockets.js'; // yarn add uWebSockets.js@uNetworking/uWebSockets.js#<tag>
+import cookie from 'cookie';
+import { makeBehavior, Request } from 'graphql-ws/lib/use/uWebSockets';
+import { schema } from './my-graphql-schema';
+import { validate } from './my-auth';
+
+// your custom auth
+class Forbidden extends Error {}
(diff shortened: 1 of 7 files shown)


maxpain:
@enisdenjo

enisdenjo:
Hey hey, I saw this - makes sense, thank you. I just want to do some more cleanup here before this lands

Will be done today, sorry for the delay, am really busy.

README.md
```diff
@@ -769,6 +769,57 @@ wsServer.on('connection', (socket, request) => {
 
 </details>
 
+<details id="uws-auth-handling">
+<summary><a href="#uws-auth-handling">🔗</a> Server usage with <a href="https://github.com/uNetworking/uWebSockets.js">uWebSockets.js</a> and custom auth handling</summary>
```
enisdenjo, line 773: I think this example is redundant (since we have the exact same thing for `ws`). Besides, you cannot set custom WebSocket upgrade headers so it is always preferred to do the auth through `connectionParams`.

src/use/uWebSockets.ts
```diff
@@ -10,12 +11,16 @@ export interface Extra {
   /**
    * The actual socket connection between the server and the client.
    */
-  readonly socket: uWS.WebSocket;
+  readonly socket: uWS.WebSocket & { upgradeReq: Request };
```
enisdenjo, line 14: Why include the upgrade request here too? The `request` below _is_ the upgrade request.
```suggestion
  readonly socket: uWS.WebSocket;
```

src/use/uWebSockets.ts
```diff
@@ -10,12 +11,16 @@ export interface Extra {
...
   /**
    * The initial HTTP request before the actual
    * socket and connection is established.
    */
-  readonly request: uWS.HttpRequest;
+  readonly request: Request;
+}
+
+export interface Request {
+  headers: http.IncomingHttpHeaders;
```
enisdenjo, line 23: If you remove the `upgradeReq` from the `socket`, this can extended inline.
```suggestion
  readonly request: Request & { headers: http.IncomingHttpHeaders };
}
```

enisdenjo (changes requested):
If the upgrade request is completely unusable after the `upgrade` callback, we can just drop it altogether in the context extra. The 2 fields being then: `socket` and `requestHeaders`.

README.md
maxpain: > you cannot set custom WebSocket upgrade headers from the browser 
Yes, but we use cookies to auth WebSocket upgrade requests, for example.

src/use/uWebSockets.ts
maxpain: It was there before my changes, I just wanted to save consistency between actual JS objects and TypeScript defenitions.

src/use/uWebSockets.ts
maxpain: Request is our own Interface with only one field: `headers`

maxpain:
> If the upgrade request is completely unusable after the `upgrade` callback, we can just drop it altogether in the context extra. The 2 fields being then: `socket` and `requestHeaders`.

I just wanted to save consistency between `ws` and uWebSocket servers.

README.md
enisdenjo: Sure, but there is already an exact example over at `ws`.

src/use/uWebSockets.ts
enisdenjo: No it was not. You added the `upgradeReq` haha, check the diff. 

Permalink from master:
https://github.com/enisdenjo/graphql-ws/blob/2933997cfbbe9965a52acadb0a727e6dfb87b265/src/use/uWebSockets.ts#L13

enisdenjo:
Understood, but the reality is that uWS and ws are quite different.  Nevertheless, changing the context extra will result in a breaking change - so, lets leave it like this for now.

src/use/uWebSockets.ts
maxpain: > No it was not. You added the `upgradeReq` haha, check the diff.

I mean it was as JS object, but not as TypeScript definition.

https://github.com/enisdenjo/graphql-ws/blob/master/src/use/uWebSockets.ts#L75

enisdenjo: I see, but the TS at uWS claim that this:
https://github.com/enisdenjo/graphql-ws/blob/2933997cfbbe9965a52acadb0a727e6dfb87b265/src/use/uWebSockets.ts#L84
is only `uWS.WebSocket`.

[You also had to assert the typing `const socket = args[0] as uWS.WebSocket & { upgradeReq: Request }`](https://github.com/maxpain/graphql-ws/blob/f97cfb4f2b33f63efcb9a72c0ee1c3b7fdac705b/src/use/uWebSockets.ts#L96)

README.md
maxpain: Ok, you can delete this example if you want.

src/use/uWebSockets.ts
maxpain: Yes, for some reason developer of `uWS` doesn't give us ability to extend `uWS.WebSocket` with extra fields, passed to upgrade function :(

maxpain: Yes, for some reason developer of `uWS` doesn't give us the ability to extend `uWS.WebSocket` with extra fields, passed to upgrade function :(

So I extended it by myself here:

https://github.com/maxpain/graphql-ws/blob/f97cfb4f2b33f63efcb9a72c0ee1c3b7fdac705b/src/use/uWebSockets.ts#L96

src/use/uWebSockets.ts
maxpain: Ok. What I need to do? In your suggested change you deleted my `Request` interface but still used as type of `request` field

enisdenjo (approved):

enisdenjo:
@maxpain made a few modifications and merged this. Took the liberty of updating the PR's description just so that future dwellers understand what changed more accurately.

Thanks for bringing this up!

maxpain:
@enisdenjo thank you

PR #228, opened 2021-09-07
feat(client): Add `connectionAckWaitTimeout` option

D34THWINGS:
On the server side exists a `connectionInitWaitTimeout` that would close the connection if no init is received but it doesn't exist the other way around. It could happen that the socket connects to an instance that is overloaded or shutting down and is not responding to messages or at least super slowly. In that case, the client should also timeout and try to reconnect and retry connection initialisation.

I added the same logic there's on the sevrer side but with a `connectionAckWaitTimeout` parameter. A 0/Infinity/null/empty value will not trigger the timeout. A positive number value will trigger a timeout that will close the socket with a specific error code and message. The timeout clears upon receiving the connection ack itself  Defaults to 0 to prevent breaking changes.

### DIFF
--- a/docs/enums/common.CloseCode.md
+++ b/docs/enums/common.CloseCode.md
@@ -11,6 +11,7 @@
 ### Enumeration members
 
 - [BadRequest](common.CloseCode.md#badrequest)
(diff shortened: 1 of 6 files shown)


enisdenjo (approved):
Thanks for this!

PR #276, opened 2021-11-30
refactor(client): add hint when missing WebSocket impl for Node

benjie:
Might save someone a few seconds :wink:


enisdenjo (approved):
LGTM, thanks Benjie!

PR #321, opened 2022-02-26
Fix behaviour around closing the socket if the onConnect handler throws

airhorns:
The JSDoc comment for the onConnect handler outlines behaviour that makes sense to me -- if the context can't be set up, we close the socket and send an error to the client. I couldn't find anything that actually implemented that though, so I added the behaviour. My test was failing before I added this change, so I think maybe the behaviour was just lost at some point.


airhorns:
Hm, I may have misunderstood exactly where this error handling is supposed to be as it seems like the tests for the `use` handlers are breaking. My project is using a custom mount of the `makeServer` object as opposed to one of the helpers from `use` so we can layer in some custom stuff of our own. Is there a way to push the `onConnect` error handling into `makeServer` instead of each binding having to recreate it?

enisdenjo:
The core server functionality has no opinions about error handling, and should never have. Close codes within `makeServer` are protocol-centric; meaning, it uses those specified in the [GraphQL over WebSocket Protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) and some relevant to it (like the [`CloseCode.SubprotocolNotAcceptable`](https://github.com/enisdenjo/graphql-ws/blob/8ef68b57f309881cd822879b77e74d68c580d365/src/server.ts#L566-L570)).

Everything in the `use/` folder is an implementation of the Protocol and therefore adds additional complexity and handles common exceptions, like throwing errors in unlikely places. (This is why there are tests failing now.)

The two are separate worlds, `makeServer` is ultra bare including just the Protocol logic, while `use/*` is much more complex and real-world ready. Combine this with the unopinionated error handling - this is why you don't see any try/catch statements in `makeServer` (literally). ["Server usage with ws" recipe](https://github.com/enisdenjo/graphql-ws#ws) might be helpful for visualising.

Anyone implementing `graphql-ws` through `makeServer` should use the `onMessage` callback argument for catching and handling all possible errors. [This is possible because the complete execution is wrapped inside `onMessage` (with the exception of the `onClose` hook if the subprotocol is invalid).](https://github.com/enisdenjo/graphql-ws/blob/8ef68b57f309881cd822879b77e74d68c580d365/src/server.ts#L590-L837) Straight from the [`onMessage` docs](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/server.WebSocket.md#onmessage):

> All operations requested from the client will block the promise until completed, this means that the callback will not resolve until all subscription events have been emitted (or until the client has completed the stream), or until the query/mutation resolves.
>
> Exceptions raised during any phase of operation processing will reject the callback's promise, catch them and communicate them to your clients however you wish.

TL;DR: wrapping quite literally all hooks that the server offers with try/catch blocks (as you did in this PR) is equivalent to:
```ts
import { makeServer } from 'graphql-ws';

const server = makeServer({ schema });

const closed = server.opened(
  {
    ...
    onMessage: (cb) => {
      ...
      socket.on('message', async (event) => {
        try {
          // wait for the the operation to complete
          // - if init message, waits for connect
          // - if query/mutation, waits for result
          // - if subscription, waits for complete
          await cb(event.toString());
        } catch (err) {
          // all errors that could be thrown during the
          // execution of operations will be caught here
          socket.close(CloseCode.InternalServerError, err.message);
        }
      });
    },
    ...
  },
  ...
);
```

With this in mind, polluting the core with try/catch blocks (if `onConnect` has it, why not every other hook?) and forcing an `InternalServerError` close code on _all_ users is a no-go. Sorry, but I hope you understand.

P.S. Those remarks you're mentioning in the `onConnect` (on throw, close) are leftovers from before, when `graphql-ws` tightly integrated with `ws` and had no `makeServer` ([pre v2.0.0](https://github.com/enisdenjo/graphql-ws/compare/v1.14.0...v2.0.0)). They're still completely true if using anything from `use/*`, but not true for the pure `makeServer`; maybe we should drop them from the docs instead? 🤔

PR #361, opened 2022-05-12
Add a hint if you use grqphql-ws in jest test

Samox:


enisdenjo:
This information is extraneous. There is already a ["Client usage in Node" recipe](https://github.com/enisdenjo/graphql-ws#node-client) and the error message when a WebSocket implementation is missing is super explanatory:
https://github.com/enisdenjo/graphql-ws/blob/9ba15d0373d0e28f02e043d9f112358e85bfd845/src/client.ts#L532-L534.

PR #375, opened 2022-06-30
fix: add support for TypeScript from an ESM project

jeevcat:
In a ESM project using TypeScript, I am unable to import types correctly.

I have followed the suggestions recommended here: https://stackoverflow.com/questions/72457791/typescript-packages-that-ship-with-mjs-and-d-ts-but-without-d-mts-how-to-i


enisdenjo (approved):
Great stuff, thanks!

PR #400, opened 2022-09-14
feat: cache for parsing and validating

mezykr:


enisdenjo:
Thanks for this! However, a cache like this is something that can be implemented in user-land.

I want to keep the core as minimal as possible, so everything that can be built on top should not be included in the core. Less code = less maintenance.

You're very welcome to push a PR where you add this cache implementation as a recipe in the readme!

PR #406, opened 2022-09-21
fix: reorder types in package.json exports

shahyar:
This puts the `"types"` of every export in `package.json` to be the first entry in the object. Without this, import resolution fails in `nodenext` and `node16` modes.


enisdenjo (approved):
Thanks!

PR #447, opened 2023-02-14
fix: ensure asyncIterator call return() when fail at calling next()

niceandneat:
This calls `.return()` of asyncIterator before remove it from `subscriptions`.

After https://github.com/microsoft/TypeScript/pull/51297, error at `iterator.next()` in `for await (... of iterator)` statement does not call `iterator.return()`. This can cause asyncIterator not to be cleared and hang.

### DIFF
--- a/src/__tests__/server.ts
+++ b/src/__tests__/server.ts
@@ -1740,6 +1740,44 @@ describe('Subscribe', () => {
       fail("Shouldn't have received a message");
     }, 20);
   });
+
+  it('should execute "return" of the subscription when error occurs in "next" call', async (done) => {
+    const iterator = (async function* () {
+      yield { data: { greetings: 'Hi' } };
+      throw new Error('error');
+    })();
+
+    jest.spyOn(iterator, 'return').mockImplementation(done);
+
+    const { url } = await startTServer({
+      schema,
+      subscribe: () => iterator,
+    });
+
+    const client = await createTClient(url);
+    client.ws.send(
+      stringifyMessage<MessageType.ConnectionInit>({
+        type: MessageType.ConnectionInit,
+      }),
+    );
+
+    await client.waitForMessage(({ data }) => {
+      expect(parseMessage(data).type).toBe(MessageType.ConnectionAck);
+      client.ws.send(
+        stringifyMessage<MessageType.Subscribe>({
+          id: '1',
+          type: MessageType.Subscribe,
+          payload: {
+            operationName: 'Greetings',
+            query: `subscription Greetings {
+                greetings
+              }`,
+            variables: {},
+          },
+        }),
+      );
+    });
+  });
 });
 
 describe('Disconnect/close', () => {
--- a/src/server.ts
+++ b/src/server.ts
@@ -824,6 +824,10 @@ export function makeServer<
               // completed the subscription, he doesnt need to be reminded
               await emit.complete(id in ctx.subscriptions);
             } finally {
+              // before delete the subcription, we should call `return` of asyncIterator to prevent it from hanging
+              const sub = ctx.subscriptions[id];
+              if (isAsyncGenerator(sub)) await sub.return(undefined);
+
               // whatever happens to the subscription, we finally want to get rid of the reservation
               delete ctx.subscriptions[id];
             }


enisdenjo (commented):
Hey thanks for this! I am just wondering why calling `.return` is necessary here. If you throw from a running iterator it is destroyed anyways, a `.return` call would be superfluous - right?

niceandneat:
@enisdenjo 

Nice point! I thought it would be necessary cause some custom async iterator implementations contain clear logic at `.return()`, or `.throw()` (like [this one](https://github.com/apollographql/graphql-subscriptions/blob/master/src/pubsub-async-iterator.ts#L59)). But if an error throws at `.next()` call during `for await` loop, neither `.return()`, nor `.throw()` runs. So it can make some external resources remain in the memory and not free them.

It can be seen as the responsibility of implementations, but this problem can be prevented by the consumer of async iterator and this PR is about that prevention.

enisdenjo:
I understand your point, but beware that neither `.return` or `.throw` are callbacks - they are methods used for controlling the iterator itself and are therefore not a good place for side-effects. Meaning, the convention of using `.return` for cleanups is actually off-putting.

As per MDN:

- [`.throw` method of a generator acts as if a throw statement is inserted in the generator's body at the current suspended position](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator/return)
- [`.return` method of a generator acts as if a return statement is inserted in the generator's body at the current suspended position](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator/return)

And then `.return` method documentation on MDN continues:

> ...which finishes the generator and allows the generator to perform any cleanup tasks when combined with a `try...finally` block

[_MDN about `.return`_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator/return)

It hints that cleanups are to be done in the `finally` block of a try-catch statement.

niceandneat:
> I understand your point, but beware that neither .return or .throw are callbacks - they are methods used for controlling the iterator itself and are therefore not a good place for side-effects. Meaning, the convention of using .return for cleanups is actually off-putting.

Yes, I also thought `.return()` should not be used for cleanups. Actually this is about restoring the old behavior.

https://github.com/enisdenjo/graphql-ws/blob/611c223471ed3aeee55a80e12d4cdc1a86176b15/src/server.ts#L847-L854

https://github.com/enisdenjo/graphql-ws/blob/611c223471ed3aeee55a80e12d4cdc1a86176b15/src/use/ws.ts#L137-L151

Before https://github.com/enisdenjo/graphql-ws/commit/611c223471ed3aeee55a80e12d4cdc1a86176b15, `.return()` for remain subscriptions was called in the return function of `.opened()` at situations like early exits(e.g. throw at `.next()`), because socket closes at `catch` block of `.opened()` call.

After that commit, thrown async iterator is removed from `ctx.subscriptions` and its `.return()` cannot be called. But other async iterators that are contained in the same subscriptions call `.return()` same as before.

enisdenjo:
We only call `.return` on async iterators if we're the ones invoking the cleanup (connection closes or user-triggered completes), otherwise we should not.

I don't fully follow your last comment, how's this behavior different with https://github.com/enisdenjo/graphql-ws/commit/611c223471ed3aeee55a80e12d4cdc1a86176b15?

A further flaw with this PR is also that `.return` will be called twice when the generator completes, this would definitely be breaking.

We must assume that if the async generator read loop exits/throws - its due to user intervention (either by calling `.return` or `.throw` or throwing in `.next`).

niceandneat:
> I don't fully follow your last comment, how's this behavior different with https://github.com/enisdenjo/graphql-ws/commit/611c223471ed3aeee55a80e12d4cdc1a86176b15?

Sorry for confusing. In short,

**Before https://github.com/enisdenjo/graphql-ws/commit/611c223471ed3aeee55a80e12d4cdc1a86176b15**

If throwing in `.next` happens at one async iterator, every async iterators in the same `ctx.subscriptions` call `.return` **includes** thrown async iterator.

**After https://github.com/enisdenjo/graphql-ws/commit/611c223471ed3aeee55a80e12d4cdc1a86176b15**

If throwing in `.next` happens at one async iterator, every async iterators in the same `ctx.subscriptions` call `.return` **excepts** thrown async iterator. Because thrown async iterator removed from `ctx.subscriptions` before socket close callback is called.

> A further flaw with this PR is also that `.return` will be called twice when the generator completes, this would definitely be breaking.

That's the side effect I missed, thanks for checking! I think this PR should be closed because of the flaw above 😅.

enisdenjo:
> If throwing in .next happens at one async iterator, every async iterators in the same ctx.subscriptions call .return excepts thrown async iterator. Because thrown async iterator removed from ctx.subscriptions before socket close callback is called.

And this is how it should be. 😅

niceandneat:
Yep I get it. Thanks for answering!

PR #448, opened 2023-02-15
docs: modernise nodejs example

fungiboletus:
NodeJS provides a randomUUID method since the version 14.17 and 15.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -877,20 +877,13 @@ if (supported) {
 
 ```ts
 const ws = require('ws'); // yarn add ws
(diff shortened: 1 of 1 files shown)


enisdenjo (approved):
Thank you! Even though [the package supports Node `>=10`](https://github.com/enisdenjo/graphql-ws/blob/eea18ce208abed12d3a9e61adc2c28d618cb3803/package.json#L28), we'll still merge this for brevity and 14 being LTS.

PR #457, opened 2023-03-06
Add information about rejecting a connection.

### LINKED ISSUE #455: graphql-transport-ws rejection specification
### Story

As a _server implementor_ I want to have a standard way to be able to **positively reject** a connection when I receive a `connection_init` message, 
so that a _client_, upon receiving that message, can know that it has been rejected, and receive some message
to that effect.

### Acceptance criteria

- _server_ is able to send a _standard_ response in response to `connection_init` indicating a rejected connection and the reason for this, followed by closing the websocket connection.
- _client_ can wait for either the usual `connection_ack`, or this new standard response, before attempting a `subscribe` message
- The protocol specification _specifies_ how to reject a connection upon receiving `connection_ack`.

### Discussion

Currently the only _in protocol_ way to not accept a connection is by _not sending_ a `connection_ack` message.  However, this is not a _positive rejection_ and a _client_ cannot know for how long he should be waiting for the `connection_ack` message.  Further, there is no way to convey a _reason_, such as "rejected", "busy", "unauthorized", etc.

Simply closing the websocket is not ideal, since that is a separate protocol layer.

### Suggestion

- We could add an _optional_ field to `connection_ack`, `success:bool`, defaulting to  `true`, but with a `false` value indicating reject.
- We could permit an `error` message to be sent at this point.  However, `error` messages are related to individual _operations_ and have an `operation-id`.  Further, their cargo is a `GraphQLError` list, which is not what this error is specifying.
- We could specify that the standard way is to close the _websocket_ with some specific error code.

kristjanvalur:
Closes #455 

EDIT by @enisdenjo: [Use the "closes" syntax instead.](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -34,7 +34,7 @@ The server must receive the connection initialisation message within the allowed
 
 If the server receives more than one `ConnectionInit` message at any given time, the server will close the socket with the event `4429: Too many initialisation requests`.
 
-If the server wishes to reject the connection, for example after examining any tokens in the `payload`, it is recommended to close the socket with the event `4403: Forbidden`.
+If the server wishes to reject the connection, for example during authentication, it is recommended to close the socket with `4403: Forbidden`.
 
 ```typescript
 interface ConnectionInitMessage {
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -34,6 +34,9 @@ The server must receive the connection initialisation message within the allowed
 
 If the server receives more than one `ConnectionInit` message at any given time, the server will close the socket with the event `4429: Too many initialisation requests`.
 
+If the server wishes to reject the connection, for example after examining any tokens in the `payload`, it should
+close the socket with the event `4403: Forbidden`.
+
 ```typescript
 interface ConnectionInitMessage {
   type: 'connection_init';
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -34,8 +34,7 @@ The server must receive the connection initialisation message within the allowed
 
 If the server receives more than one `ConnectionInit` message at any given time, the server will close the socket with the event `4429: Too many initialisation requests`.
 
-If the server wishes to reject the connection, for example after examining any tokens in the `payload`, it should
-close the socket with the event `4403: Forbidden`.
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -34,6 +34,9 @@ The server must receive the connection initialisation message within the allowed
 
 If the server receives more than one `ConnectionInit` message at any given time, the server will close the socket with the event `4429: Too many initialisation requests`.
 
+If the server wishes to reject the connection, for example after examining any tokens in the `payload`, it should
+close the socket with the event `4403: Forbidden`.
```
enisdenjo, line 38: Don't add new lines for sentences.

```suggestion
If the server wishes to reject the connection, for example after examining any tokens in the `payload`, it should close the socket with the event `4403: Forbidden`.
```

enisdenjo, line 38: "should" is a strong word, we don't want to require a 4403 ever - this is just a recommendation.

```suggestion
If the server wishes to reject the connection attempt, it is recommended to close the socket with `4403: Forbidden`.
```

enisdenjo (changes requested):

PROTOCOL.md
```diff
@@ -34,6 +34,8 @@ The server must receive the connection initialisation message within the allowed
 
 If the server receives more than one `ConnectionInit` message at any given time, the server will close the socket with the event `4429: Too many initialisation requests`.
 
+If the server wishes to reject the connection, for example after examining any tokens in the `payload`, it is recommended to close the socket with the event `4403: Forbidden`.
```
enisdenjo, line 37: Less specific example.

```suggestion
If the server wishes to reject the connection, for example during authentication, it is recommended to close the socket with `4403: Forbidden`.
```

enisdenjo (changes requested):

PROTOCOL.md
kristjanvalur: sure.

kristjanvalur:
Great, thanks!

PR #466, opened 2023-03-10
Base server (class based)

ilijaNL:
Websocket implementation based on ws-graphql protocol without depending on graphql as discussed in https://github.com/enisdenjo/graphql-ws/discussions/461

TODO:
- [ ] Documentation
- [ ] Test for base-server
- [ ] Different use-cases/implementations

### DIFF
--- a/src/base-server.ts
+++ b/src/base-server.ts
@@ -1,0 +1,622 @@
+/**
+ *
+ * server
+ *
+ */
+import {
+  CloseCode,
+  ID,
+  Message,
+  MessageType,
+  stringifyMessage,
+  parseMessage,
+  ConnectionInitMessage,
+  SubscribeMessage,
+  NextMessage,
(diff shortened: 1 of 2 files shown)


enisdenjo (commented):
Oh, I thought you wanted to create a separate library - not have this abstraction here. This is a GQL WS server and it should stay as such without any unnecessary complications or abstractions.

Sorry about any misunderstanding.

ilijaNL:
> 

Yes it is perhaps better, also for leaking (graphql) implementation. However I wonder what you think about this idea/implementation

enisdenjo:
The code is MIT licensed for a reason, I'd love to see what you can make out of it. 😄

ilijaNL:
See https://github.com/enisdenjo/graphql-ws/discussions/461 for follow up

PR #473, opened 2023-05-08
fix(ws): wait for subscriptions to finish when closing server

ttoohey:
The `dispose()` method would resolve before a subscription that has cleanup tasks in a `return()` method has completed. This change allows those subscriptions to complete so that awaiting `dispose()` will ensure that all subscription cleanup tasks have completed


ttoohey:
closing as https://github.com/enisdenjo/graphql-ws/pull/474 should allow for cleanup to be performed without having side effects in `.return()`

PR #474, opened 2023-05-08
fix(server): wait for async generator cleanup to complete

ttoohey:
allow subscriptions to have a cleanup phase after calling `.return()` and ensure the iterator is 'done' before finishing

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -847,7 +847,12 @@ export function makeServer<
       return async (code, reason) => {
         if (connectionInitWait) clearTimeout(connectionInitWait);
         for (const sub of Object.values(ctx.subscriptions)) {
-          if (isAsyncGenerator(sub)) await sub.return(undefined);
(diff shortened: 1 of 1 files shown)


enisdenjo:
Invoking the `return` method of an iterator should finish it immediately. Except for wrong iterator implementations - I don't think there are cases where you need to flush pending messages after completing it.

> `return(value)`
> A function that accepts zero or one argument and returns an object conforming to the `IteratorResult` interface, typically with `value` equal to the `value` passed in and `done` equal to `true`. Calling this method tells the iterator that the caller does not intend to make any more `next()` calls and can perform any cleanup actions.

[_*The iterator protocol on MDN_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol)

Having read the `return` method definition on MDN, we can notice 2 things about it:
- `done` should be `true`
- `next()` shouldn't be called after `return()`

PR #507, opened 2023-10-04
Add retry info to connect events

telemenar:
This adds retry info to the connecting and connected events. This helps us give the user better status about the current state of their network connection with the websocket.


enisdenjo (commented):
This looks good! Thanks, would you be so kind and add some tests? Just in case.

PR #511, opened 2023-10-10
GraphQL-WS x Nextra 3

B2o5T:


website/public/sitemap.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
enisdenjo, line 1: Like with `graphql-sse`, would be great if you can gitignore this file since its generated on each deploy. Thanks!

enisdenjo (changes requested):
🥳

PR #512, opened 2023-10-11
Fix close event code for BadGateway

thumbsupep:
This corrects the close event code for BadGateway from 1013 to 1014 https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code#value 

This error is considered non fatal and users should be able to retry.

### DIFF
--- a/src/__tests__/client.ts
+++ b/src/__tests__/client.ts
@@ -1695,6 +1695,7 @@ describe('reconnecting', () => {
     console.warn = warn;
     await testCloseCode(CloseCode.InternalServerError);
     await testCloseCode(CloseCode.InternalClientError);
+    await testCloseCode(CloseCode.BadGateway);
     await testCloseCode(CloseCode.BadRequest);
     await testCloseCode(CloseCode.BadResponse);
     await testCloseCode(CloseCode.Unauthorized);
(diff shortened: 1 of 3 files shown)


src/common.ts
```diff
@@ -29,6 +29,7 @@ export const DEPRECATED_GRAPHQL_WS_PROTOCOL = 'graphql-ws';
 export enum CloseCode {
   InternalServerError = 4500,
   InternalClientError = 4005,
+  BadGateway = 1014,
```
enisdenjo, line 32: These are application specific close codes, close codes that are used by the GraphQL over WS spec.

```suggestion
```

src/__tests__/client.ts
```diff
@@ -1695,6 +1695,7 @@ describe('reconnecting', () => {
     console.warn = warn;
     await testCloseCode(CloseCode.InternalServerError);
     await testCloseCode(CloseCode.InternalClientError);
+    await testCloseCode(CloseCode.BadGateway);
```
enisdenjo, line 1698: This does not belong here, the test states "should report some close events immediately and not reconnect". The client _should_ reconnect on a 1014.

```suggestion
```

enisdenjo (changes requested):
A few remarks, otherwise looks good!

src/__tests__/client.ts
thumbsupep: just acknowledging that I read this! it might take me a bit to write a unit test because I'm used to mocking apollo-specific web client servers and am only working on this on weekends 😅

enisdenjo: If you're busy - I can finish this. The required changes are ultra minimal - basically removing code.

enisdenjo (approved):

PR #522, opened 2023-11-29
Improve randomized exp backoff algorithm

iblislin:


iblislin:
performance, I think.
oh, okay, so the readability is more critically?

PR #523, opened 2023-11-29
docs(protocol): clarify id field is not optional

roksui:
This PR aims to clarify the section of the protocol which states that the `id` field is optional even though this field is either required or not present as per the type definitions. (Also discussed in [#340](https://github.com/enisdenjo/graphql-ws/discussions/340#discussioncomment-7703620))

At first, I wanted to mention specifically that `Subscribe`, `Next`, `Error`, and `Complete` message types needs to have the `id` field. However, the sentence that needs to be clarified is in the "Communication" section which comes before the explanation of the message types. So, I used the term "Messages corresponding to operations" to refer to those message types. 

Though, I am not sure if this sentence sounds clear enough. I am happy to accept any improvements for wording.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -11,10 +11,9 @@ The WebSocket sub-protocol for this specification is: `graphql-transport-ws`.
 
 Messages are represented through the JSON structure and are stringified before being sent over the network. They are bidirectional, meaning both the server and the client must conform to the specified message structure.
 
-**All** messages contain the `type` field outlining the action this message describes. Depending on the type, the message can contain two more _optional_ fields:
+**All** messages contain the `type` field outlining the action this message describes. Depending on the type, the message can contain the _optional_ field `payload` holding the extra "payload" information to go with the specific message type.
 
-- `id` used for uniquely identifying server responses and connecting them with the client's requests
-- `payload` holding the extra "payload" information to go with the specific message type
+Messages corresponding to operations contain the `id` field used for uniquely identifying server responses and connecting them with the client's requests.
 
 Multiple operations identified with separate IDs can be active at any time and their messages can be interleaved on the connection.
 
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -11,10 +11,9 @@ The WebSocket sub-protocol for this specification is: `graphql-transport-ws`.
 
 Messages are represented through the JSON structure and are stringified before being sent over the network. They are bidirectional, meaning both the server and the client must conform to the specified message structure.
 
-**All** messages contain the `type` field outlining the action this message describes. Depending on the type, the message can contain two more _optional_ fields:
+**All** messages contain the `type` field outlining the action this message describes. Depending on the type, the message can contain the _optional_ field `payload` holding the extra "payload" information to go with the specific message type.
```
enisdenjo, line 14: The `payload` is also not always optional. I would just leave out the whole next sentence.

```suggestion
**All** messages contain the `type` field outlining the action this message describes.
```

PROTOCOL.md
```diff
@@ -11,10 +11,9 @@ The WebSocket sub-protocol for this specification is: `graphql-transport-ws`.
...
 
-- `id` used for uniquely identifying server responses and connecting them with the client's requests
-- `payload` holding the extra "payload" information to go with the specific message type
+Messages corresponding to operations contain the `id` field used for uniquely identifying server responses and connecting them with the client's requests.
```
enisdenjo, line 16:
```suggestion
Messages corresponding to operations must contain the `id` field used for uniquely identifying server responses and connecting them with the client's requests.
```

enisdenjo:
Thank you very much!

roksui:
@enisdenjo Thank you! Should I be concerned about the failing check for this PR?

enisdenjo:
Nope, it's just Deno acting up (it happens from time to time). 😅

PR #538, opened 2024-02-16
Expand Error Handling Test Coverage

divanise:
This PR adds tests covering how errors from the `subscribe` and `resolve` functions associated with a field definition are handled. The commit messages go into the details in each commit, but as a simplified overview, the test GraphQL schema has been updated to include a query and a subscription that allows the caller to control whether an error is thrown from various parts of their implementations. Using that, the client and server tests have been updated to compare the results of various permutations of those operations against snapshots of the current behaviour.

This PR makes no changes outside of test code; I've added these tests to both learn about and document the existing behaviour rather than to fix a specific bug.


divanise:
I've run out of time to follow up on this, so I'm closing this for now.

PR #573, opened 2024-08-04
fix(server): Close code and reason are optional for `close`

tobiasdiez:
Follow-up to https://github.com/enisdenjo/graphql-ws/commit/6ae6e6f4e392975ad786cf61e4947fd57f5cf56c, which missed the `close` method


enisdenjo (approved):
Great, thanks!

PR #585, opened 2024-10-25
Fix WS types, current ones are broken

ruizmarc:
WS library has changed how types are exported and graphql-ws is breaking builds if type checking for libraries is active:

```
node_modules/graphql-ws/lib/use/ws.d.ts(5,28): error TS2339: Property 'prototype' does not exist on type 'typeof import("/ptp/node_modules/@types/ws/index")'.

node_modules/graphql-ws/lib/use/ws.d.ts(6,27): error TS2694: Namespace '"/ptp/node_modules/@types/ws/index"' has no exported member 'Server'.
```

I have just properly imported types so we can use the library without problems. 

Let me know if you need anything else from me. Hope to see it released soon!

### DIFF
--- a/src/use/ws.ts
+++ b/src/use/ws.ts
@@ -1,5 +1,5 @@
 import type * as http from 'http';
-import type * as ws from 'ws';
+import type { WebSocket, WebSocketServer } from 'ws';
 import { handleProtocols, makeServer, ServerOptions } from '../server';
 import {
   DEPRECATED_GRAPHQL_WS_PROTOCOL,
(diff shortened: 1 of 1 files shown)


src/use/ws.ts
```diff
@@ -9,10 +9,6 @@ import {
 } from '../common';
 import { limitCloseReason } from '../utils';
 
-// for nicer documentation
-type WebSocket = typeof ws.prototype;
-type WebSocketServer = ws.Server;
-
```
enisdenjo, line 15: These should be put back because they're used for documentation.

```suggestion
// for nicer documentation
export type { WebSocket, WebSocketServer };
```

enisdenjo (changes requested):
Thanks for the update! One small change request.

Also, does this work for older versions for `ws`? It wont be a breaking change? Furthermore, why aren't typechecks detecting this - does the `@types/ws` package need an update?

src/use/ws.ts
ruizmarc: Done :)

enisdenjo (approved):

PR #599, opened 2024-11-26
fix(server): onError should return GraphQLFormattedError

benjie:
The types sent over the wire should be _formatted_ types - specifically errors should be piped through `formatError` and the types shouldn't imply that they're `instanceof GraphQLError` - they're just data.

> [!WARNING]
> This is a breaking change _if_ anyone's `onError` callback is relying on `instanceof GraphQLError` checks (seems extremely unlikely?), but otherwise it should be safe because `GraphQLError` conforms to `GraphQLFormattedError`.

> [!TIP]
> I don't actually care about `formatError` being called in core (feel free to revert that), I just care that the type of the `onError` hook allows a formatted error be returned for consistency, and that the payload type should not imply the errors are `instanceof GraphQLError`.

> [!NOTE]
> The payloads that send `ExecutionResult` should also be sending `FormattedExecutionResult` but that seemed like a larger change so I didn't make it.


enisdenjo (approved):

benjie (commented):
Looks good :+1:

PR #621, opened 2025-01-21
fix(common): Fix the `FormattedExecutionResult.errors` type

pleunv:
https://github.com/enisdenjo/graphql-ws/releases/tag/v6.0.0 introduced a `FormattedExecutionResult`, however, its `errors` type seems to be incorrect as it's set to `ReadonlyArray<FormattedExecutionResult>` rather than `ReadonlyArray<GraphQLFormattedError>`:

```ts
/** @category Common */
export interface FormattedExecutionResult<
  Data = Record<string, unknown>,
  Extensions = Record<string, unknown>,
> {
  errors?: ReadonlyArray<FormattedExecutionResult> | undefined;
  data?: Data | null | undefined;
  hasNext?: boolean | undefined;
  extensions?: Extensions | undefined;
}
```

This is remediated by this PR.

### DIFF
--- a/.changeset/eleven-bears-warn.md
+++ b/.changeset/eleven-bears-warn.md
@@ -1,0 +1,5 @@
+---
+'graphql-ws': patch
+---
+
+FormattedExecutionResult errors field returns GraphQLFormattedError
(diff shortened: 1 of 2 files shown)


enisdenjo (approved):
Oof what an oversight... Thanks!

PR #625, opened 2025-02-12
Use Math.pow for retry delay calculation

HermanBilous:
Simplified the logic for default retryWait function

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -464,14 +464,11 @@ export function createClient<
     connectionAckWaitTimeout = 0,
     retryAttempts = 5,
     retryWait = async function randomisedExponentialBackoff(retries) {
-      let retryDelay = 1000; // start with 1s delay
-      for (let i = 0; i < retries; i++) {
-        retryDelay *= 2;
-      }
+      const retryDelaySeconds = Math.pow(2, retries);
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -464,14 +464,11 @@ export function createClient<
     connectionAckWaitTimeout = 0,
     retryAttempts = 5,
     retryWait = async function randomisedExponentialBackoff(retries) {
-      let retryDelay = 1000; // start with 1s delay
-      for (let i = 0; i < retries; i++) {
-        retryDelay *= 2;
-      }
+      const retryDelaySeconds = Math.pow(2, retries);
```
enisdenjo, line 467: If it's meant to be "seconds" (because of the suffix), then:
```suggestion
      const retryDelaySeconds = Math.pow(2, retries) * 1000;
```
Right?

enisdenjo:
If you would like to release a new version following this change, please run `yarn changeset` and follow the instructions. Thanks!

src/client.ts
enisdenjo: Oof, right. Great!

enisdenjo (approved):
Thanks, this is looking good!

PR #652, opened 2025-08-28
WIP: allow custom protocols to be included in the opening connection

GingerGeek:
I am looking to introduce the ability to define custom protocols in the websocket opening so that servers can use this for authentication or other custom data.

This is because that under Browser WebSocket APIs you cannot include custom headers in the HTTP connection that gets upgraded (e.g. `Authorization`)

GraphQL-ws library already suggests two authentication points. One is by encoding a token in the URL `wss://myserver.example.com/graphql?token=<>` and the other is by utilising the `ConnectionInit` and opening phase of the protocol.

The use of URL paramaters for authentication is generally discouraged due to the risk of leaking into proxy, network or system logs. ConnectionInit would generally preferred but is not available in all implementations.

Here is a documented use of a major project using subprotocols to enable connection-time authentication for generic websocket connections: https://quarkus.io/guides/websockets-next-reference#bearer-token-authentication

As it happens, my colleague is also attempting to add support into Quarkus to support authentication at ConnectionInit within the context of graphql-ws https://github.com/quarkusio/quarkus/pull/49643/

This PR is not yet ready for formal review and is missing tests, but I'm just looking for a comment on if this feature would be allowed.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -226,6 +226,18 @@ export interface ClientOptions<
    */
   connectionParams?: P | (() => Promise<P> | P);
   /**
+   * Optional protocols, passed through to the WebSocket constructor that will be part of 
+   * the `Sec-WebSocket-Protocol` header sent to the server. As browser WebSocket connections
+   * do not allow setting arbitary headings this can be used by server implementations to send
+   * arbitary data for example Authentication headers, especially when it's desired to happen 
+   * prior to a WebSocket conncetion being answer
+   * 
+   * If the option is a function, it will be called on every WebSocket connection attempt.
+   * Returning a promise is supported too and the connecting phase will stall until it
+   * resolves with the URL.
+   */
+  additionalProtocols?: string[] | (() => Promise<string[]> | string[]);
+  /**
    * Controls when should the connection be established.
    *
    * - `false`: Establish a connection immediately. Use `onNonLazyError` to handle errors.
@@ -456,6 +468,7 @@ export function createClient<
   const {
     url,
     connectionParams,
+    additionalProtocols,
     lazy = true,
     onNonLazyError = console.error,
     lazyCloseTimeout: lazyCloseTimeoutMs = 0,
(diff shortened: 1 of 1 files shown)


enisdenjo:
Also remember the 3rd, often overlooked, option: cookies. You will always have them accompany the initial HTTP upgrade request. In my personal opinion, [they should always be used for session auth in the browser](http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/).

> jwt-flowchart
>
> _[Slightly sarcastic flowchart by Sven Slootweg](http://cryto.net/%7Ejoepie91/blog/2016/06/19/stop-using-jwt-for-sessions-part-2-why-your-solution-doesnt-work/)_

---

Back to the topic here; while using subprotocols for token transfer is clever, I cannot condone it. [That is not what the Sec-WebSocket-Protocol header is designed for.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Protocol) Since it's an anti-pattern, I am not keen on advertising it as an option in this repo and therefore spreading misuse. GraphQL over WebSocket is a subprotocol that should be used with graphql-ws.

Nevertheless, you can still provide a custom WebSocket implementation and move the token into the subprotocol list, something like:

```ts
class URLToSubprotocolTokenWebSocket extends WebSocket {
  constructor(address, protocols) {
    const url = new URL(address);
    const token = url.searchParams.get('token'); // extract
    url.searchParams.remove('token'); // remove
    super(url, [...protocols, token]);
  }
}

const client = createClient({
  url: 'ws://localhost:4000/graphql?token=<your-token>',
  webSocketImpl: URLToSubprotocolTokenWebSocket,
});
```

I really appreciate your contribution and insights though! If you want, you can add this example to the website recipe and document it a bit more. 🙏

GingerGeek:
Yeah, I agree it's cursed. Thank you for such a prompt reply.

For me, misusing Sec-WebSocket-Protocol is a lesser evil than putting JWTs into query, but I empathise with your reservation on promoting it.

Cookies are not an easy option for us, the vast majority of connections to our GraphQL endpoints are by non-browser/app environments, and our authentication model there is based on Oauth service accounts (for better or worse). Plus the auth server, web app and graphql endpoint all sit on different domains 😅 

As mentioned in the original PR our preferred solution is to have Quarkus's implementation of graphql-ws protocol support authentication.

We have some reservations that this allows too much code path to run prior to an authentication check which could open up to DDoS related concerns but we will monitor.

One of the challenges we got from Quarkus maintainers is that there is no specification for what should be used in ConnectionInit headers for authentication, I know this is probably a deliberate choice on your side to keep the protocol to a minimum but having some suggestions in the website may help alternative server/client implementations conform uniformly.

I will try and find some time to contribute to the website to discuss this

enisdenjo:
Thank you for taking the time here! Very much looking forward to discussing this publically on the website.

Nevertheless, my [example of `URLToSubprotocolTokenWebSocket` implementation](https://github.com/enisdenjo/graphql-ws/pull/652#issuecomment-3232833938) is completely valid and as such should unblock you.

PR #655, opened 2025-08-31
Update example to remove AsyncGenerator (#654)

monitorjbl:
AsyncGenerators are not safe to use outside of toy examples. The `next()` function will block any call to `return()` or `throw()` if it is waiting on a promise. To avoid this pitfall, implementations should not use the native `async function*()` syntax and should instead provide their own `AsyncIterator` implementations.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -1,0 +1,3 @@
+# README
+
+Building the website requires that the documentation be generated from the root project into the `src/pages/docs` directory _first_. This can be done by `yarn gendocs` from the root of the repo.
--- a/website/src/pages/get-started.mdx
+++ b/website/src/pages/get-started.mdx
@@ -16,6 +16,7 @@ npm i graphql-ws
 
 ```ts
 import { GraphQLObjectType, GraphQLSchema, GraphQLString } from 'graphql';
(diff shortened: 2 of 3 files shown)


enisdenjo (changes requested):
Thanks for working on this! However, I feel like the 50 lines of `EventProvider` for the getting started is too much. I instead recommend using [RepeaterJS](https://repeater.js.org/) here.

If you want to talk about the asyngenerator quirks, maybe add it somewhere else? A new page? Recipes?

monitorjbl:
> Thanks for working on this! However, I feel like the 50 lines of `EventProvider` for the getting started is too much. I instead recommend using [RepeaterJS](https://repeater.js.org/) here.
> 
> If you want to talk about the asyngenerator quirks, maybe add it somewhere else? A new page? Recipes?

Sure, moved the longer explanation to recipes and made the getting started example use Repeater.js

monitorjbl:
Any other feedback, or is this good to merge?

enisdenjo (approved):
Sorry for the delay, this is looking good. Let's wait for the CI to pass and merge. Thank you for your support!

PR #664, opened 2026-01-07
Update recipes.mdx abruptly closed example

loucadufault:
the client can reconnect even if the close code is 1000:

https://github.com/enisdenjo/graphql-ws/blob/0c0eb499c3a0278c6d9cc799064f22c5d24d2f60/src/client.ts#L827-L830


enisdenjo (changes requested):
There are some CI issues, please fix the checks. Otherwise looks good, thanks!

enisdenjo (approved):

PR #667, opened 2026-03-17
fix: do not send Complete after Error for subscriptions

endigma:
Per the protocol spec:

> **Error:** This message terminates the operation and no further messages will be sent.

> **Complete (Server → Client):** If the server dispatched the `Error` message relative to the original `Subscribe` message, no `Complete` message will be emitted.

The server was sending both messages because the subscription ID remained in `ctx.subscriptions` after `emit.error()`, causing `emit.complete()` to notify the client.

The fix deletes the subscription immediately after emitting the error, so the existing guard at `emit.complete()` skips the notification. A new test verifies the protocol compliance.

### DIFF
--- a/.changeset/khaki-buses-trade.md
+++ b/.changeset/khaki-buses-trade.md
@@ -1,0 +1,26 @@
+---
+'graphql-ws': patch
+---
+
+Fix the server sending a `Complete` message after an `Error` message for subscriptions.
+
+Previously, when a subscription's async iterable threw an error, the server would send:
+
+```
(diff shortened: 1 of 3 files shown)


enisdenjo (changes requested):
This is great work, thank you for catching it! Do you mind writing a changeset for it so that we can release ASAP? `yarn changeset` will kick you off.

endigma:
Hi @enisdenjo, thanks for taking a look! I added a changeset, let me know if there's any issues with it.

enisdenjo (approved):

PR #680, opened 2026-07-22
fix(client): don't remove an unrelated listener when disposing of an already-removed listener

kkhys:
The unsubscribe function returned by `client.on` removes the listener with `l.splice(l.indexOf(listener), 1)`. When the listener has already been removed, `indexOf` yields `-1` and `splice(-1, 1)` removes the last listener of the same
event instead of doing nothing.

This corrupts the listener registry without any double-dispose by the user: emits iterate over a copy of the listeners, so when `client.terminate()` is called from within a `closed` listener, the nested synthetic `closed` emit makes the internal one-shot `errorOrClosed` listeners unlisten themselves — and the still-running outer emit re-invokes them from its stale copy, so they unlisten again and knock registered `closed`/`error` listeners out of the registry. From that point on the client silently stops reporting `closed`/`error` events.

We hit this in production: a client that terminates on a 4401 close (to reconnect with a fresh token) lost its `closed` listener this way, leaving the app blind to later disconnects for the rest of a call.

Fixed by guarding the unsubscribe with an `indexOf` check so that removing an already-removed listener is a no-op. Added two regression tests that fail on master.

### DIFF
--- a/.changeset/afraid-pandas-repeat.md
+++ b/.changeset/afraid-pandas-repeat.md
@@ -1,0 +1,7 @@
+---
+'graphql-ws': patch
+---
+
+Disposing of a client event listener twice no longer removes an unrelated listener
+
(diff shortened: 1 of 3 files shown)


enisdenjo (approved):
excellent work! thank you very much

PR #682, opened 2026-07-24
feat(server): Add `parse` option for custom GraphQL parsing

jwatzman:
We allow swapping out `validate`, `execute`, etc, and should allow `parse` too. This is useful for caching, logging, and the like.

### DIFF
--- a/.changeset/gold-students-roll.md
+++ b/.changeset/gold-students-roll.md
@@ -1,0 +1,5 @@
+---
+'graphql-ws': minor
+---
+
+Add `parse` option for custom GraphQL parsing
--- a/src/server.ts
+++ b/src/server.ts
@@ -11,11 +11,11 @@ import {
   GraphQLError,
   execute as graphqlExecute,
   GraphQLFormattedError,
+  parse as graphqlParse,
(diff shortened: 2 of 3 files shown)


enisdenjo (changes requested):
this is looking good, thanks! please format the code by running `yarn format` before we merge

jwatzman:
Whoops! Fixed.

enisdenjo (changes requested):
one more thing is left, please run `yarn changeset` and document your change so that we can release a new version after this PR lands

jwatzman:
@enisdenjo happy to do so. Would this be a `patch` or `minor`? I think by the letter of semver it would be `minor` since it's a new feature (sorta?), but it's such a small change I know some folks prefer `patch` for that sort of thing. Which would you prefer here?

enisdenjo:
it's a new feature, so I'd go with a minor bump

jwatzman:
OK, I think/hope I did that right? Not super familiar with changesets.

enisdenjo (approved):
that's great, thanks!

PR #686, opened 2026-09-11
fix(crossws): close the socket even if server.opened closes immediately

### LINKED ISSUE #669: CrossWS support cannot close WebSocket in server.opened
The guard in here https://github.com/enisdenjo/graphql-ws/blob/fc0300468644ea117142bc94adbda5d79181828b/src/use/crossws.ts#L73 prevents the server from closing the WebSocket inside the opened function (e.g. here https://github.com/enisdenjo/graphql-ws/blob/fc0300468644ea117142bc94adbda5d79181828b/src/server.ts#L590) as the peer is only added to the clients set upon returning from opened.

The result is that no init response is sent and the WebSocket stays open.

cpruijsen:
## Summary

The CrossWS adapter defers `peer.close` with `setImmediate`, so a close issued from `server.opened` reaches the socket.

`close`/`send` no-op unless the peer is in a WeakMap that is only populated after `opened` returns. A close from `opened` (for example the 4406 protocol-mismatch close) was ignored, the WebSocket stayed open, and no `ConnectionAck` was sent because the message handler was never installed.

The `setImmediate` deferral matches the uWebSockets adapter (`src/use/uWebSockets.ts`, "end socket in next tick making sure the client is registered"). Registering the peer in the WeakMap before `server.opened` is not enough if `peer.close` is synchronous (the close hook would still see the stub closure handler). Alternative: `queueMicrotask` for non-Node CrossWS runtimes. Can switch to `queueMicrotask` for Workers/Deno.

Reported by @andreasdamm-shure in #669 with the exact guard and the `opened` close path.

A client that offers no subprotocol must close with 4406 / "Subprotocol not acceptable" on every adapter, including CrossWS.

Closes #669

## Test plan

- [x] `tests/use.test.ts` new test "should close the socket if the client does not offer an acceptable subprotocol" fails on CrossWS without the fix (times out: socket stays open) and passes with it; `ws` / uWebSockets / `@fastify/websocket` already passed
- [x] Full `tests/use.test.ts` (all adapters)


enisdenjo (approved):
brilliant. lets wait for the CI to pass and we can land this - thanks!

PR #691, opened 2026-09-16
feat(server): add onPing and onPong callbacks to ServerOptions

### LINKED ISSUE #650: Expose onPing and OnPong, similar to onConnect, onDisconnect.......
I have a requirement where i have to log the subprotocol level ping/pong received.

I have seen the recipe for it, it works but.... feels offf

can we provide onPing and onPong, similar to onConnect, ... like a method with access to context...

niukanen1:
## Summary

Closes #650

Adds `onPing` and `onPong` callbacks to `ServerOptions`, similar to `onConnect` / `onDisconnect`. Both receive the connection `Context` as the first argument and the message `payload` as the second, so apps can log subprotocol-level pings with access to connection state

The automatic pong reply is preserved — unlike the low-level `socket.onPing` listener which disables it for manual control. If the low-level listener is registered, the server-level callback runs first and then control is handed over

Work with or without connection acknowledgement, like the message handling itself

## Testing

`yarn vitest run`: 262 passed, 0 failed


src/server.ts
```diff
@@ -669,9 +706,13 @@ export function makeServer<
...
           }
           case MessageType.Ping: {
-            if (socket.onPing)
+            if (socket.onPing) {
               // if the onPing listener is registered, automatic pong is disabled
+              await onPing?.(ctx, message.payload);
               return await socket.onPing(message.payload);
+            }
+
+            await onPing?.(ctx, message.payload);
```
enisdenjo: one onPing on the top would be good enough, right?

```suggestion
            await onPing?.(ctx, message.payload);

            if (socket.onPing) {
              // if the onPing listener is registered, automatic pong is disabled
              return await socket.onPing(message.payload);
            }
```

enisdenjo (changes requested):
one micro nit, otherwise gtg

niukanen1:
Applied in 37315a1 — the callback now runs once at the top of the Ping case

enisdenjo (approved):
much appreciated, great work

